PR #29, opened 2022-10-20
feat: switched to unstorage

### LINKED ISSUE #28: use cache mechanism from `unstorage`
Nitro has native caching built-in. It implements this using `unstorage`. You could hook into unstorage also for caching instead of using `memory-cache`. It is as simple as calling `useStorage()` which is exposed by Nitro.

Intevel:
Will close #28


Intevel:
@danielroe @Baroshem

Baroshem (approved):

danielroe (commented):
It might be worth adding a namespace prefix to the IP rather than directly using it

Issue #30, opened 2022-10-21
Some module improvements


pi0:
Doing a quick review for https://github.com/nuxt/modules/pull/473

- (refactor/enhancenment) A suggested by @danielroe in https://github.com/Baroshem/nuxt-security/issues/27 most of rules could be written with new routeRules instead of middleware
- Use `throw createError()` (https://github.com/Baroshem/nuxt-security/blob/main/src/runtime/server/middleware/requestSizeLimiter.ts#L19)
- (perf) You can cache XSSFilter instance once (https://github.com/Baroshem/nuxt-security/blob/main/src/runtime/server/middleware/xssValidator.ts#L10)
- (bug) storage getItem/setItem are async and should be awaited on (https://github.com/Baroshem/nuxt-security/blob/29a07e70786e25a1abc818d5a3978241137a4a1f/src/runtime/server/middleware/rateLimiter.ts#L12) it probably doesn't works. 
- (docs) In the headers page (https://nuxt-security.vercel.app/middlewares/headers) it would be nice if you can also document actual headers are being added for transparency

Baroshem:
1. Works well
2. Implemented

PR #32, opened 2022-10-21
fix: use throw createError instead of sendError

Intevel:
use throw createError instead of sendError, as mentioned in #30 from @pi0

## Types of changes


Baroshem:
As we talked on Discord, I have to close it for different reasons. Thanks for the help!

PR #43, opened 2022-11-03
Update nitro.ts

Maxp777:
## Types of changes

## Checklist:

### DIFF
--- a/src/runtime/nitro.ts
+++ b/src/runtime/nitro.ts
@@ -1,7 +1,7 @@
 import type { NitroAppPlugin, RenderResponse } from 'nitropack'
 
-export default <NitroAppPlugin> function (nitro) {
-  nitro.hooks.hook('render:response', (response: RenderResponse) => {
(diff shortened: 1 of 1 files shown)


Baroshem:
Hey @Maxp777 

Could you explain why this change is needed?

Maxp777:
like here https://nitro.unjs.io/guide/advanced/plugins

Baroshem:
@Maxp777 

But it works the same as what it is there right now. Also, this is how other modules are registering nitro plugins -> https://github.com/nuxt-modules/html-validator/blob/main/src/runtime/nitro.ts

Maxp777:
> @Maxp777
> 
> But it works the same as what it is there right now. Also, this is how other modules are registering nitro plugins -> https://github.com/nuxt-modules/html-validator/blob/main/src/runtime/nitro.ts

I don't understand why they added it then defineNitroPlugin

PR #65, opened 2022-11-27
feat: improve headers config and types

Qrzy:
## Types of changes

## Description

This change aims to:
- update the misleading `| boolean` in types where `true` is actually an unacceptable value that would break the module
- provide better typing for certain headers' values
- enable modifying only picked parts of complex headers like Content Security Policy

Especially the very last point pushed me towards proposing this change as I found it really cumbersome to copy, modify and paste the original value for CSP when trying to add [Nuxt Image](https://v1.image.nuxtjs.org/get-started/) using Cloudinary.

Change should be fully backwards-compatible (see using the string value when it's given instead of an object) - if there is any breaking change - I'll definitely rethink it and try to improve this proposal.

## Checklist:

- [ ] I have added tests to cover my changes (if not applicable, please state why) - well, there are no test in the repo 🙃

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -11,12 +11,11 @@ import {
   ModuleOptions,
   RateLimiter,
   RequestSizeLimiter,
-  SecurityHeader,
   SecurityHeaders,
   XssValidator
 } from './types'
 import { defaultSecurityConfig } from './defaultConfig'
-import { SECURITY_HEADER_NAMES } from './headers'
+import { SECURITY_HEADER_NAMES, getHeaderValueFromOptions } from './headers'
 
 declare module '@nuxt/schema' {
   interface NuxtOptions {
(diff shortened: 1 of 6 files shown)


src/module.ts
```diff
@@ -26,40 +26,46 @@ export default defineNuxtModule<ModuleOptions>({
...
-          nuxt.options.nitro.routeRules!![(nuxt.options.security.headers as SecurityHeader)[header].route] = { headers: nuxt.options.nitro.routeRules!![(nuxt.options.security.headers as SecurityHeader)[header].route] ? { ...nuxt.options.nitro.routeRules!![(nuxt.options.security.headers as SecurityHeader)[header].route].headers } : {} }
-
-          nuxt.options.nitro.routeRules!![(nuxt.options.security.headers as SecurityHeader)[header].route].headers!![SECURITY_HEADER_NAMES[header]] = (nuxt.options.security.headers as SecurityHeader)[header].value
+    if (securityOptions.headers) {
+      for (const header in securityOptions.headers as SecurityHeaders) {
+        const headerOptions = (securityOptions.headers as SecurityHeader)[header]
+        if (headerOptions) {
+          const nitroRouteRules = nuxt.options.nitro.routeRules
+          nitroRouteRules!![headerOptions.route] = {
+            ...nitroRouteRules!![headerOptions.route],
```
Qrzy, line 50: Also, added this little fella, as I found the previous version disregarding any potential previous of entire route rules.

Baroshem:
@Qrzy, I think that I have rebased the code in a wrong way (as it was a lot of files to adjust). Could you please revert it so that I can adjust it to the current main?

Sorry for the trouble.

Qrzy:
@Baroshem, this should be ready for review now :)

Baroshem:
Thanks @Qrzy !

I will merge it today and begin testing!

PR #80, opened 2022-12-29
fix: use custom merger for options to avoid array values concatenation

### LINKED ISSUE #78: adding custom configuration with array structure results in duplication of header values
When registering a custom configuration like for CSP headers with the array structure as instructed in the docs, user will get a duplicated headers values. This is caused by `defu` package that merges two objects and if they dont find exact match, it will merge two objects together. 

I will instruct in the docs to not use this approach as it will cause bugs.

### Version

nuxt-security: 
nuxt:

Qrzy:
## Types of changes

## Description

As pointed in #78, `defu` (both used here and in Nuxt's code) was the reason of even triplicated values in the headers.
This PR aims fixing that by applying 2 changes:
- removing defaults from module definition - so that Nuxt does not apply regular `defu` on defaults and options from the module user
- creating custom merger with `createDefu` that in case of arrays replaces the value instead of concatenating them

Fixes #78

## Checklist:

### DIFF
--- a/playground/nuxt.config.ts
+++ b/playground/nuxt.config.ts
@@ -31,4 +31,21 @@ export default defineNuxtConfig({
   //     route: '/upload-file'
   //   }
   // }
+  security: {
+    headers: {
+      contentSecurityPolicy: {
+        value: {
+          'img-src': ["'self'", 'data:', 'https://dummy.test']
+        },
+        route: '/**'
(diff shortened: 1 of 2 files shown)


Baroshem (approved):

Baroshem:
Works great! Merging.

PR #102, opened 2023-02-16
Fix incorrect auth condition in basic-auth

Tristan971:
## Types of changes

## Description

The condition to check for interruption of basic auth should be one of:
- No credentials
- Invalid credentials

Before the former condition was checking if both are true at the same time, which isn't correct.

Also removed the async marker on the event handler since there's no async work being done.

It could also be done with a split of the check into 2 steps, returning a 401 on missing credentials and a 403 on invalid credentials (that would be closer to what the HTTP spec says about those status codes) like so:

```ts
if (!credentials) {
  setHeader(event, 'WWW-Authenticate', `Basic realm=${basicAuthConfig.message || 'Please enter username and password'}`)
  sendError(event, createError({ statusCode: 401, statusMessage: 'Access denied' }))
}
if (!validateCredentials(credentials!, basicAuthConfig)) {
  setHeader(event, 'WWW-Authenticate', `Basic realm=${basicAuthConfig.message || 'Invalid username and/or password'}`)
  sendError(event, createError({ statusCode: 403, statusMessage: 'Access denied' }))
}
```

## Checklist:

No test added mainly because I don't see any test setup in the project and am unfamiliar with the best practices for that with nuxt modules.

### DIFF
--- a/src/runtime/server/middleware/basicAuth.ts
+++ b/src/runtime/server/middleware/basicAuth.ts
@@ -1,5 +1,5 @@
-import { defineEventHandler, setHeader, createError, sendError } from 'h3'
 import getCredentials from 'basic-auth'
+import { createError, defineEventHandler, sendError, setHeader } from 'h3'
 import { useRuntimeConfig } from '#imports'
 
 type Credentials = {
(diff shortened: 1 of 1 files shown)


Tristan971:
Applied it @Baroshem 👍

Baroshem:
Thank you @Tristan971 

Welcome to the contributors!

If you have any other ideas on how we can improve the module together, please let me know :)

Issue #136, opened 2023-04-01
Right script-src parameter in Content-Security-Policy


alexbidenko1998:
## Problem description

At the moment, Content-Security-Policy header do not include `script-src` parameter. It means, that every script can be executives in user browser - and it is not secured. As a minimal solution I can propose: `'script-src': ["'self'", "'unsafe-inline'"]` - it will be blocking `eval` and unsecured scripts from other recourses, but it is not right.

More secured solution is do not use `unsafe-inline`. But for this `nuxt-data` must be resolved by means of `nonce` or `sha-***`. But I do not know how do it.

### Describe the solution you'd like

In ideal world `script-src` should include `'sha-***'` of `nuxt-data` and automatically calculate it in every html rendering on server side.

As alternative `script-src` can include `'nonce-***'` and the nonce must be injected as an attribute in `nuxt-data` `script` (for better security `nonce` must be randomize on every request, do not be constant).

### Additional context

In Next.js framework we just can add `<NextData nonce="next-data" />` component and inject `nonce-next-data` in `script-src` - it is simple correct solution.

Baroshem:
Hey,

Thank you for reporting this issue! I will take a look at it in the upcoming days :)

Baroshem:
I just talked about this during Nuxt Insiders meeting and we came up with some ideas on how this can be implemented. I will be working on it next week preferably :)

Baroshem:
Hey, not yet. 

I am working on it. Have been quite busy recently with other topics.

If anyone would like to contribute to it, feel free to open a PR with the change :)

Baroshem:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src

Baroshem:
Hey, not really.

Recently, I did not have time to work on this feature. It is open for contributions for anyone who would like to tackle this. I will be able to come back to this earliest next week or the week after

trijpstra-fourlights:
I've implemented nonce support for CSP in #171 

You can test this directly from github by using:
`
 "nuxt-security": "github:trijpstra-fourlights/nuxt-security#build/nonce",
`
in your package.json.

Then in your `nuxt.config.ts` add:

```
security: {
  // ...
  nonce: true,    // <-- add to enable nonce
  headers: {
      contentSecurityPolicy: {
        // ... other csp settings
        'script-src': [
          "'self'", // backwards compatibility for older browsers that don't support strict-dynamic
[...]

PR #171, opened 2023-07-11
feat: add `nonce` support to csp

trijpstra-fourlights:
## Types of changes

## Description

Add support for `nonce` values in CSP. 
Regenerates nonce on every request, but allows overriding this behaviour per route. 

I'm using this in production to implement a [strict csp using nonce](https://web.dev/strict-csp/#what-is-a-strict-content-security-policy).

Resolves: #136 

## Checklist:

### DIFF
--- a/src/runtime/nitro/plugins/cspNonce.ts
+++ b/src/runtime/nitro/plugins/cspNonce.ts
@@ -15,6 +15,12 @@ interface NuxtRenderHTMLContext {
   bodyAppend: string[]
 }
 
+// To prevent the nonce attribute from being added to literal strings,
+// we need to make sure that the tag is not preceded by a single or double quote.
+// This is done by using a negative lookbehind assertion. See https://www.regular-expressions.info/lookaround.html
+// See https://regex101.com/r/DBE57j/1 for some examples.
+const tagNotPrecededByQuotes = (tag: string) => new RegExp(`(?<!['|"])<${tag}`, 'g')
+
 export default <NitroAppPlugin> function (nitro) {
   nitro.hooks.hook('render:html', (html: NuxtRenderHTMLContext, { event }: { event: H3Event }) => {
     const nonce = parseNonce(`${event.node.res.getHeader('Content-Security-Policy')}`)
@@ -22,12 +28,12 @@ export default <NitroAppPlugin> function (nitro) {
     if (!nonce) { return }
 
     // Add nonce attribute to all link tags
-    html.head = html.head.map(link => link.replaceAll(/<link/g, `<link nonce="${nonce}"`))
-    html.bodyAppend = html.bodyAppend.map(link => link.replaceAll(/<link/g, `<link nonce="${nonce}"`))
+    html.head = html.head.map(link => link.replaceAll(tagNotPrecededByQuotes('link'), `<link nonce="${nonce}"`))
+    html.bodyAppend = html.bodyAppend.map(link => link.replaceAll(tagNotPrecededByQuotes('link'), `<link nonce="${nonce}"`))
 
     // Add nonce attribute to all script tags
-    html.head = html.head.map(script => script.replaceAll(/<script/g, `<script nonce="${nonce}"`))
-    html.bodyAppend = html.bodyAppend.map(script => script.replaceAll(/<script/g, `<script nonce="${nonce}"`))
+    html.head = html.head.map(script => script.replaceAll(tagNotPrecededByQuotes('script'), `<script nonce="${nonce}"`))
+    html.bodyAppend = html.bodyAppend.map(script => script.replaceAll(tagNotPrecededByQuotes('script'), `<script nonce="${nonce}"`))
   })
 
   function parseNonce (content: string) {
--- a/src/runtime/nitro/plugins/cspNonce.ts
+++ b/src/runtime/nitro/plugins/cspNonce.ts
@@ -1,0 +1,41 @@
+import type { NitroAppPlugin } from 'nitropack'
+import type { H3Event } from 'h3'
+import type {
+  ModuleOptions
+} from '../../../types'
+import { useRuntimeConfig } from '#imports'
+
+interface NuxtRenderHTMLContext {
+  island?: boolean
+  htmlAttrs: string[]
+  head: string[]
+  bodyAttrs: string[]
+  bodyPrepend: string[]
+  body: string[]
+  bodyAppend: string[]
+}
+
+export default <NitroAppPlugin> function (nitro) {
+  nitro.hooks.hook('render:html', (html: NuxtRenderHTMLContext, { event }: { event: H3Event }) => {
+    const nonce = parseNonce(`${event.node.res.getHeader('Content-Security-Policy')}`)
+
+    if (!nonce) { return }
+
+    // Add nonce attribute to all link tags
+    html.head = html.head.map(link => link.replaceAll(/<link/g, `<link nonce="${nonce}"`))
+    html.bodyAppend = html.bodyAppend.map(link => link.replaceAll(/<link/g, `<link nonce="${nonce}"`))
+
+    // Add nonce attribute to all script tags
+    html.head = html.head.map(script => script.replaceAll(/<script/g, `<script nonce="${nonce}"`))
+    html.bodyAppend = html.bodyAppend.map(script => script.replaceAll(/<script/g, `<script nonce="${nonce}"`))
+  })
+
+  function parseNonce (content: string) {
+    const noncePattern = /nonce-([a-zA-Z0-9+/=]+)/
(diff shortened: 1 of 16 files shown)


Baroshem:
Hey @trijpstra-fourlights 

Thank you so much for this pull request!

I have been so busy recently that I wasn't able to work on this highly requested feature :(

I will review it in the upcoming days and probably release a new version on Friday with this feature 🚀

src/runtime/nitro/plugins/cspNonce.ts
```diff
@@ -0,0 +1,41 @@
...
+    const nonce = parseNonce(`${event.node.res.getHeader('Content-Security-Policy')}`)
+
+    if (!nonce) { return }
+
+    // Add nonce attribute to all link tags
+    html.head = html.head.map(link => link.replaceAll(/<link/g, `<link nonce="${nonce}"`))
+    html.bodyAppend = html.bodyAppend.map(link => link.replaceAll(/<link/g, `<link nonce="${nonce}"`))
+
+    // Add nonce attribute to all script tags
+    html.head = html.head.map(script => script.replaceAll(/<script/g, `<script nonce="${nonce}"`))
```
alexbidenko1998, line 29: What about something like this:

```html
<script>var myVar = '<script>'</script>
```

I mean if `<script` text will be contained inside script. If I do not mistake, evety value in html.head (and others) array is single script or link element. May be regex should be `/^<link/g` (with `^` symbol).

src/runtime/server/middleware/cspNonceHandler.ts
```diff
@@ -0,0 +1,46 @@
...
+        if (!nonce) {
+          return sendError(event, createError({ statusCode: 401, statusMessage: 'Nonce is not set' }))
+        }
+
+        break
+      }
+      case 'renew':
+      default: {
+        nonce = nonceConfig?.value ? nonceConfig.value() : Buffer.from(crypto.randomUUID()).toString('base64')
+        setCookie(event, 'nonce', nonce, { sameSite: true, secure: true })
```
alexbidenko1998, line 33: What about if site will not use https connection? Maybe it would be better to look at strictTransportSecurity option in NuxtSecurity config?

src/runtime/server/middleware/cspNonceHandler.ts
```diff
@@ -0,0 +1,46 @@
...
+  const routeRules = getRouteRules(event)
+
+  if (routeRules.security.nonce !== false) {
+    const nonceConfig: NonceOptions = routeRules.security.nonce
+
+    // See if we are checking the nonce against the current value, or if we are renewing the nonce value
+    let nonce: string | undefined
+    switch (nonceConfig?.mode) {
+      case 'check': {
+        nonce = getCookie(event, 'nonce')
```
alexbidenko1998, line 22: Is it secured to use cookie for nonce value? I do not know, sorry, but what do you think about https://nuxt.com/docs/api/composables/use-nuxt-data#usenuxtdata , for example?

src/types.ts
```diff
@@ -90,6 +96,7 @@ export type CSPSourceValue =
   | "'none'"
   | "'strict-dynamic'"
   | "'report-sample'"
+  | "'nonce=<base64-value>'"
```
alexbidenko1998, line 99: Is it corrected for type? May be this: `'nonce=${string}'`

src/runtime/server/middleware/cspNonceHandler.ts
trijpstra-fourlights: As far as I know, using a cookie with `sameSite` and `secure` will only allow access to that cookie from scripts which are already loaded, i.e. which adhere to the nonce value.

I tried using useNuxtData and similar, but there seems no added benefit to security, and the implementation becomes more complex as you need to add specific endpoints for retrieving the nonce value (and how is it going to persist between regeneration?)

Furthermore, `@sidebase/nuxt-auth` uses a similar approach to pass tokens to the browser as well. It's also recommended by the nuxt documentation.

I'll see if I can find the correct links for this.

src/types.ts
trijpstra-fourlights: It's generally suggested to use a base64 encoded random string.
The current implementation does that, but it's not a hard requirement per se.

src/runtime/server/middleware/cspNonceHandler.ts
alexbidenko1998: Won't it interfere with development mode (with localhost without https)? Some alert should be shown, I think, because it is not transparent for understanding in currect sulution (or looking at strictTransportSecurity option).

src/runtime/nitro/plugins/cspNonce.ts
trijpstra-fourlights: I was unaware of this property of `html.head`!
If it's indeed guaranteed that every value in `html.head` is a single element then your suggestion makes a lot of sense.

A more simple approach would be to just use `replace` instead of `replaceAll` in that case. Probably a tiny bit quicker than using a regex.

alexbidenko1998: One of examples in this discussion comment: https://github.com/nuxt/nuxt/issues/18376#issuecomment-1438303970

Can it help you better?

src/runtime/server/middleware/cspNonceHandler.ts
trijpstra-fourlights: To be honest, I'm hoping that any users of this library will not be serving a webapp without SSL in this day and age.

I would even then suggest to just add a middleware to this plugin which checks for:

* if production and not on ssl -> throw warning about security without SSL being pointless

That being said, it seems that secure: true cookies are currently transmitted on localhost domains (probably to developers).

src/runtime/nitro/plugins/cspNonce.ts
trijpstra-fourlights: I'm not sure I see the problem with the current implementation.

Your example of :

```
<script>var myVar = '<script>'</script>
```

would not be allowed when using nonce, unless you also add `strict-dynamic`  to CSP.
Otherwise, it will need a `nonce` attribute in any case, as it otherwise won't be adhering to the CSP policy.

The current implementation will add the nonce to that inside script tag as well, which would work with or without `strict-dynamic`.

I personally would wait with adding extra parsing logic (and thus increasing code complexity) until it shows that people are having problems with the chosen approach.

Still, you are of course free to add or change this PR if you feel that's necessary in this case.

src/runtime/server/middleware/cspNonceHandler.ts
alexbidenko1998: I agree with you that ssl is must have))

But "if production and not on ssl" - if site will work inside docker container and proxy without ssl - checking will failed too.

I think `secure` might be with true value, as you have done)

src/runtime/server/middleware/cspNonceHandler.ts
alexbidenko1998: I have not think about non ssr variant... Now cookies is looked like very good solution))

src/runtime/nitro/plugins/cspNonce.ts
alexbidenko1998: In my example I mean that your solution also will replace string inside content of script element. But someone might to use `<script>` text for... I don not know... in text for examples in documentation on page of some site. It will be just text, but `/<script/g` regex replace and it too.

trijpstra-fourlights: Yeah I understand what you mean, but is that really a real-world scenario? We're talking about a nuxt module for a nuxt app here: someone would have to go out of their way to add a script like that in the head tag instead of using a component.

Furthermore, they then start using this nuxt-security module, and opt-in to the nonce behaviour: a conscious decision that you only do if you care about enforcing a strict csp policy. In any case they should be thinking about these scripts and why they are loading them in this non-nuxtonian-way.

I just can't fanthom that being a real world scenario.

Still, I can add a test for your specific situation and update the RegExp accordingly. IMO the solution is to check that there are no quotes (`'` or `"`) around the script tag.

trijpstra-fourlights: See https://github.com/Baroshem/nuxt-security/pull/171/commits/828387ef16b105141fbb2a7a8b4e38a72718258e

alexbidenko1998: Thank you very much 😊

My worried soul will be calm)))

Baroshem:
Hey @trijpstra-fourlights @alexbidenko1998 Thank you so much for such advanced discussion about the implementation of this feature. I have tested it locally and it looks like it is working really well. Solid work.

I wil proceed with releasing it tomorrow as the new version of NuxtSecurity.

Stay Tuned!

PR #177, opened 2023-07-17
fix: nonce quirks when using `ssr`

trijpstra-fourlights:
## Types of changes

## Description
When working with the recently introduced `nonce` feature I noticed some quirks that this PR aims to correct.

- `nonce`  attribute was not being applied to `<style>` elements, resulting in CSP violations
- when using `ssr` the nonce value could get out of sync
- using the `nonce` value directly is almost never needed, so the documentation should show that as well

Finally, I discovered strange behavior due to [an upstream bug in `unjs/unhead`](https://github.com/unjs/unhead/issues/136) when using nonce values.

- useHead tags are injected twice when using `nonce` and `ssr`

I've fixed these quirks and updated the tests and documentation accordingly to reflect these findings.

## Checklist:

### DIFF
--- a/docs/content/2.security/1.headers.md
+++ b/docs/content/2.security/1.headers.md
@@ -3,7 +3,7 @@ title: Headers
 description: ''
 ---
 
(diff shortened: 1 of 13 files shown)


Baroshem (approved):

Baroshem:
I will merge it now and release it once I will fix the CSP for SSG but it should be max tomorrow evening :)

Issue #194, opened 2023-08-09
Invalid permission policy parser


nekotoriy:
### Version

nuxt-security: 0.14.4
nuxt: 3.6.5

### Reproduction Link

[Stackblitz](https://stackblitz.com/edit/nuxt-starter-zsju5u?file=nuxt.config.ts)

### Steps to reproduce

```js
// nuxt.config.ts / security.headers
security: {
  headers: {
    permissionsPolicy: {
      geolocation: ["self", '"https://api.website.com"'],
      midi: [],
      payment: [],
      camera: ["self", '"https://api.website.com"'],
      microphone: ["self", '"https://api.website.com"'],
      usb: [],
      magnetometer: [],
      "document-domain": [],
    },
  },
},
```

### What is Expected?

```
geolocation=(self "https://api.website.com"), midi=(), payment=(), camera=(self "https://api.website.com"), microphone=(self "https://api.website.com"), usb=(), magnetometer=(), document-domain=()
```

### What is actually happening?
```
camera=self "https://api.website.com", display-capture=(), fullscreen=(), geolocation=self "https://api.website.com", microphone=self "https://api.website.com"
```

Baroshem:
Hey!

Thanks for reporting this issue!

From what I understand, this bug appears only when you are passing an array. Is that correct?

Would you be interested in fixing this bug? :)

nekotoriy:
Sure, here is [PR](https://github.com/Baroshem/nuxt-security/pull/195)

Baroshem:
Thank you so much @nekotoriy 

I will review it pater today and will release a patch version :)

PR #195, opened 2023-08-09
Added parentheses for permissions policy parser according issue

nekotoriy:
Fixes: https://github.com/Baroshem/nuxt-security/issues/194

### DIFF
--- a/src/headers.ts
+++ b/src/headers.ts
@@ -41,7 +41,15 @@ const headerValueMappers = {
     })
       .filter(Boolean).join('; ')
   },
-  permissionsPolicy: (value: PermissionsPolicyValue) => Object.entries(value).map(([directive, sources]) => (sources as string[])?.length && `${directive}=${(sources as string[]).join(' ')}`).filter(Boolean).join(', ')
+  permissionsPolicy: (value: PermissionsPolicyValue) =>
+    Object.entries(value)
+      .map(
+        ([directive, sources]) =>
+          (sources as string[])?.length &&
+          `${directive}=(${(sources as string[]).join(' ')})`
+      )
+      .filter(Boolean)
+      .join(', ')
 }
 
 export const getHeaderValueFromOptions = <T>(headerType: keyof SecurityHeaders, headerOptions: MiddlewareConfiguration<T>) => {


Baroshem:
Hey,

When I tried this PR with default configuration it does not work correctly and throws an error in the browser console 
image

Tested with following config:

```ts
    permissionsPolicy: {
      'camera': ['()'],
      'display-capture': ['()'],
      'fullscreen': ['()'],
      'geolocation': ['()'],
      'microphone': ['()'],
    }
```

nekotoriy:
@Baroshem 

By default `permissionsPolicy` already set to `()`. Is it's necessary handle overriding default values with default values ?

Baroshem:
@nekotoriy I meant that I was using the default config. 

With changes from your PR, the the module is not working correctly in terms of permissions policy. It may work for your case, but it does not work with the default config or when the user will pass just a single `['()']` as value for certain policy.

I think you would have to add some condition to the parser to only add additional parentheses when value passes is an array with multiple elements.

nekotoriy:
But in this case, passing `["self"]` will not work.

Baroshem:
Hey @nekotoriy 

sorry for no response but I was on holidays.

In this case we have two options then:
1. Change the current behavior to add parenteses automatically (it will be a Breaking Change)
2. Adjust the behavior to work for your case and for the previous approaches

From my perspective, the better would be the option two as it wont break the functionality for the current users.

However, the option one might be better for the future so we could plan it for the 1.0.0 version where we could ship breaking changes.

Let me know which option would work better from your side and we can think about it :)

nekotoriy:
Hello @Baroshem!

At the moment I don't have an urgent need to fix this bug, so we can schedule break changes for version 1.0.0 following the first point.

nekotoriy:
Alright, thank you!

Baroshem:
Hey @nekotoriy 

I have changed the target branch to 1.0.0-rc.1 as I want to include this functionality in the upcoming 1.0.0-rc.1 release.

I have tested it and it seems to be working as expected :)

src/headers.ts
```diff
@@ -41,7 +41,15 @@ const headerValueMappers = {
     })
       .filter(Boolean).join('; ')
   },
-  permissionsPolicy: (value: PermissionsPolicyValue) => Object.entries(value).map(([directive, sources]) => (sources as string[])?.length && `${directive}=${(sources as string[]).join(' ')}`).filter(Boolean).join(', ')
+  permissionsPolicy: (value: PermissionsPolicyValue) =>
+    Object.entries(value)
+      .map(
+        ([directive, sources]) =>
+          (sources as string[])?.length &&
```
Baroshem, line 48: issue: here you can safely delete the checking for length as we will allow for empty arrays that will be translated to `()`

src/headers.ts
```diff
@@ -41,7 +41,15 @@ const headerValueMappers = {
     })
       .filter(Boolean).join('; ')
   },
-  permissionsPolicy: (value: PermissionsPolicyValue) => Object.entries(value).map(([directive, sources]) => (sources as string[])?.length && `${directive}=${(sources as string[]).join(' ')}`).filter(Boolean).join(', ')
+  permissionsPolicy: (value: PermissionsPolicyValue) =>
```
Baroshem, line 44: issue: please change defaultConfig.ts value for permissionsPolicy from the existing one to:

```
    permissionsPolicy: {
      camera: [],
      'display-capture': [],
      fullscreen: [],
      geolocation: [],
      microphone: []
    }
```

This with the addition of the next comment would allow to have both () and your functionality :)

Baroshem (changes requested):

Baroshem:
Hey @nekotoriy 

Do you need any help with the implementation. I can do this if you dont have time (I will still mention you in the release notes as the contributor so no worries about that :) )

Baroshem:
@nekotoriy Are you planning to add the changes I mentioned? If you dont have time, just let me know. I will add the changes and include you in the release notes :)

Baroshem:
Closing in favor of #212

PR #213, opened 2023-09-21
fix: remove broken test for `nonce`

trijpstra-fourlights:
Removed the test, as I couldn't get it to work and it was a ridiculous one anyway. 
If you are inserting literals with `<script> etc` into your app's head you are asking for trouble.

I also had to increase the `expectedNonceElements` in the test, because `vue-export-helper` plugin is also loaded now.


Baroshem:
Damn, that was fast. Thanks! 🚀

Baroshem (approved):

PR #220, opened 2023-09-27
chore(deps): switch to pnpm

dargmuesli:
## Types of changes

- [x] Chore (some internal changes)

## Description
Hey there, most nuxt package use pnpm which makes use of symlinks and thus speeds up installations as well as it saves on disk usage, so I took the opportunity to suggest to replace yarn with pnpm. If the CI runs fine, that should count as a passed test, I'd say.

## Checklist:


dargmuesli:
I think this
https://github.com/Baroshem/nuxt-security/blob/bf5d18ec02252eeb32bedaabf062ff370b3c4726/src/module.ts#L59
should be this
```ts
const viteRemove = (await import('unplugin-remove/vite')).default
```
(notice the `default`)

---

Also, I'm not sure about the build failure, but to work around the warnings one can add this `build.config.ts`:
```
export default {
  failOnWarn: false
}
```

Baroshem:
Hey @dargmuesli 

Thank you so much for this PR but as we discussed in direct messages, there is a reason why I have selected yarn for package manager for this module.

I will close this PR now.

PR #223, opened 2023-09-29
ci: run on all pull requests and more branches

dargmuesli:
## Types of changes

- [x] CI change

## Description
This is a proposal to let the ci run on all pull requests, e.g. for this one not going against `main`.

It also proposes to let the ci run on pushes to the typical `beta` and `alpha` branches as well as branches created by the maintenance tool `renovate`, but as that's not in use currently that change can of course be removed from this PR.
Would renovate be an option for you?

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,10 +3,11 @@ name: ci
 on:
   push:
     branches:
+      - alpha
+      - beta
       - main
+      - renovate/**
   pull_request:
-    branches:
-      - main
 
 jobs:
   ci:


.github/workflows/ci.yml
```diff
@@ -3,10 +3,11 @@ name: ci
 on:
   push:
     branches:
+      - alpha
+      - beta
```
Baroshem, line 7: suggestion: I don't think that I will ever use alpha / beta. I prefer to use rc.X convention. Maybe you could add a wildcard branch here, something like `**-rc.**`

dargmuesli: Changed :raised_hands:

Baroshem (approved):

Baroshem:
Nice work @dargmuesli 💚

Issue #231, opened 2023-10-05
Runtime config not applied


becem-gharbi:
### Version

nuxt-security: 0.14.4
nuxt: 3.7.4

### Description
First, I would like to thank you for this great work!

The issue here is regarding the runtime configuration. On module setup the route rules set [here](https://github.com/Baroshem/nuxt-security/blob/0d36fec295e9b8bb2d5749d3330ff88397fe8ac6/src/module.ts#L181) are not updated at runtime. Thus replacing config options with matching environment variables ([docs](https://nuxt.com/docs/guide/going-further/runtime-config#environment-variables)) is not feasible. As a result passing variables to docker container is not feasible as the image already built with the default config options.

### Reproduction Link
https://stackblitz.com/edit/github-d931tu?file=nuxt.config.ts

### Steps to reproduce
Run application and check logs of `corsHandler` option on route rules and runtime config.

### What is Expected?
The route rules should match the runtime configuration.

### What is actually happening?
The route rules are not updated with the change in runtime configuration. The change here is the override of matching env.

becem-gharbi:
I suggest moving `setSecurityRouteRules` from module setup to a server middleware.

Baroshem:
Hey @becem-gharbi 

Thanks for this issue.

I am not sure if I understand your idea correctly.

Currently in the server middleware we are getting the values of the route rules that were set in the module configuration. I am not sure how this would work if all logic would be moved to server middleware (actually not sure if this will even work like that)

But maybe you have some idea already in mind?

If so, maybe you would be interested in creating a Proof of Concept PR with it?

The best would be to create a new branch from 1.0.0-rc.1 as I am planning to release it in the next weeks :)

becem-gharbi:
Regardless of the implementation of POC, how do you find this issue, is it a tolerated or intentional by design.

Baroshem:
Closing in favor of https://github.com/Baroshem/nuxt-security/issues/258

PR #233, opened 2023-10-06
Support runtime configuration

becem-gharbi:
## Types of changes

## Description

Solves ( #231 )

**Original behaviour**
The route rules are set at build time and cannot be changed at runtime. Thus any environment based variable such the cors `origin` cannot be passed to Docker containers and should be included in the image.

**Updated behaviour**
The route rules are created at server plugin. The `security` config object is set by path and passed on request via `event.context`.

## Checklist:
- [ ] I have added tests to cover my changes


src/module.ts
```diff
@@ -188,27 +188,6 @@ const setSecurityResponseHeaders = (nuxt: Nuxt, headers: SecurityHeaders) => {
   }
 }
 
-const setSecurityRouteRules = (nuxt: Nuxt, securityOptions: ModuleOptions) => {
```
Baroshem, line 191: question: shouldn't we also define the same approach for line 174 -> `setSecurityResponseHeaders`?

Baroshem (commented):
Some comments to address.

Also @danielroe do you think that it is a good approach to set route rules in the server plugin instead of in the module.ts file?

src/module.ts
becem-gharbi: Yes, the same.

yarn.lock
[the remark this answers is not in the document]
becem-gharbi: No dependency change.

trijpstra-fourlights (changes requested):

src/module.ts
becem-gharbi: I have made the changes to allow setting `headers` at runtime. However, objects cannot be overriden by matching environment variables. This may be related to Nuxt.

```ts
NUXT_SECURITY_HEADERS_CROSS_ORIGIN_RESOURCE_POLICY=false // Work
NUXT_SECURITY_HEADERS_CONTENT_SECURITY_POLICY=false // Do not work
```

trijpstra-fourlights (approved):
LGTM!

src/module.ts
Baroshem: If objects cannot be ovewritten, then a lot of functionality wont benefit from this change. Almost all middlewares and several headers are configured via objects (CSP, PermissionsPolicy, StrictTransportPolicy, etc).

I am not sure about this change then. In general the idea is really good, but if we are changing it and we cannot provide the same functionality as it is right now (supporting object configuration) then this can lead to bugs and issues. 

I am affraid we would need to wait for the time when Nuxt core would solve the issue you mentioned.

package-lock.json
Baroshem, line 1: For some reason I still can see package-lock.json and yarn.lock updated files even though they shouldn't be.

Could you check it out?

becem-gharbi: I have revert the changes on yarn.lock and switched to yarn.

src/module.ts
becem-gharbi: I totally agree. 

The changes here requires the ability to change route rules at runtime, which is not feasible with Nitro and the ability to override config objects with environment variables, which is not feasible with Nuxt. I hope there would be a native solutions to these in the future.

Anyway, thanks for your time and efforts on this module.

Baroshem: @becem-gharbi would you be ok with leaving this PR open for now until a proper fix will be released?

This is a great feature for sure, I just dont want to ship something that solves one issue and introduce another. Once the fix will be done, I am completely into looking at this once again and merging when ready :)

Baroshem: Also, could you create an issue in Nuxt about it and link it here? It will be easier to keep track of changes and start working on it once released :)

becem-gharbi: I have created the issue, https://github.com/nuxt/nuxt/issues/23695

Baroshem:
Hi @becem-gharbi 

Julien created a similiar PR with support for runtime configuration. Could you take a look at it and see if it solves your issue as well?

https://github.com/Baroshem/nuxt-security/issues/258

I think it will be safer to have just one implementation that solves all cases :)

becem-gharbi:
I have read its docs, indeed it's much more flexible. Thanks for the great work!

Baroshem:
Awesome, great to hear.

And sorry for long wait from my side. End of year was quite busy so I am planning right now to get back to doing much more OS work :)

If you will have any ideas on how to make the module better, please let us know :)

Issue #239, opened 2023-10-09
Nonce implementation for style src


erikknorren:
Hello,

I recently tried implementing nonce based content-security-policy from the documentation using the config listed below. The script tags in my head all have the nonce attribute and work fine this way. However, my style tags do not. In my app.vue I did the following to try to get it to work as well, however my stylesheets cannot be loaded into the DOM. Is this a bug or am I doing something wrong?

```
onMounted(() => {
  // set nonce to all style tags
  const styleTags = document.querySelectorAll('style')
  for (const styleTag of styleTags) {
    styleTag.setAttribute('nonce', useNonce())
 }
})
```

```
security: {
    nonce: true,
    headers: {
      crossOriginEmbedderPolicy: 'unsafe-none',
      contentSecurityPolicy: {
        'img-src': ['*', 'data:', 'https:'],
        'style-src': [
          "'self'", // fallback value for older browsers, automatically removed if `strict-dynamic` is supported.
          "'nonce-{{nonce}}'",
        ],
        'script-src': [
          "'self'", // fallback value for older browsers, automatically removed if `strict-dynamic` is supported.
          "'nonce-{{nonce}}'",
          "'strict-dynamic'",
        ],
        'script-src-attr': [
          "'self'", // fallback value for older browsers, automatically removed if `strict-dynamic` is supported.
          "'nonce-{{nonce}}'",
          "'strict-dynamic'",
        ],
      },
    },
  },
```

Baroshem:
Hey, thanks for rainsing a question.

I would love to get some feedback from @trijpstra-fourlights who is an expert of nonce ;)

trijpstra-fourlights:
I wouldn't call myself an expert per se ;), but I'm happy to try and help!

The `nonce` attribute cannot be read at runtime by the browser, nor set retroactively which is essentially what you're doing in the `onMounted()` snipplet.

> This is by design (as in, by the browser) because allowing that would defeat the whole purpose of the nonce: "hello browser, this specific resource is absolutely trusted by me, the server. You can check this because the `nonce` attribute is set by me and it matches the CSP header."

To get nonce working, you'll have to provide the nonce attribute when the actual <style> element is created, or use the `useHead` composable which `nuxt-security` supports out of the box.

It makes sense if you think about it: 

[...]

PR #240, opened 2023-10-12
chore: update `nonce` docs about `unsafe-inline` during development

trijpstra-fourlights:
As requested in https://github.com/Baroshem/nuxt-security/issues/239#issuecomment-1757724311


Baroshem (approved):
Amazing work. Thanks!

PR #244, opened 2023-10-16
Fix TypeScript Config

vejja:
## Types of changes
- [x] New feature: TSC linter fixed

## Description
This PR addresses the TSC configuration of the repo. It solves all linting errors, all warnings about unidenitied imports, and all unnecessary TypeScript type casts and duplicate definitions.

The code now abides by the recommendations defined by the Nuxt core team for Module Authors. 
In particular, it reflects 
- The latest module configuration recommendations (as in https://nuxt.com/docs/guide/going-further/modules#quick-start)
- The latest package versions for nuxt and related dependencies

Please use 
```
npm upgrade
npm run dev:prepare
````
to update all dependencies and regenerate type definitions before using.

## Checklist:


vejja:
@Baroshem I cannot see the vercel bot report.
Let me know if there is anything I need to do

Baroshem:
Hey @vejja 

Thank you so much for this PR.

I really appreciate your work here but we have to split some work here as there are too many changes in one PR that I can review and safely approve. Not to mention the upcoming 1.0.0-rc.1 release where I have tested all major things and with your PR merged I cannot guarantee the stability.

What I would like to recommend is to split the PR into smaller chunks. First of all, could you please extract only the fix that we talked about in the discussion? The one about nonce.

This we should be able to merge for the rc.1 version and for the rest of your PR, let's keep it for rc.2 version that will be coming probably next week or the week after.

Would this work for you?

vejja:
Ok makes sense indeed
Will do in smaller chunks

PR #248, opened 2023-10-17
Fix/typescript-config

vejja:
This PR aims at improving the TypeScript configuration of the repo

## Types of changes

## Description

Currently, the TypeScript linter emits several warnings throughout the codebase.
- We get warnings that the module "#imports" was not found, forcing us to use `@ts-ignore` overrides
- We need to reimport types from far-fetched places (such as 'h3' or 'nitropack'), where they should be available
- We need to import standard Nuxt utility functions (such as `defineEventHandler` from 'h3') from modules dependencies, i.e. that are not directly defined in our package.json
- We sometimes need to cast types to 'unknown' before re-casting them to the right type, to suppress type warnings

Solution: The Nuxt core team has formulated recommendations on how to setup the TS configuration of Nuxt modules (https://nuxt.com/docs/guide/going-further/modules).

This involves
1- setting the root tsconfig file to extend to `./.nuxt` instead of `./playground/.nuxt`
2- setting autoImports to false in `.nuxtrc`
3- setting the playground tsconfig file to extend `./.nuxt`
The configuration is detailed in the reference template for Nuxt3 modules at https://github.com/nuxt/starter/tree/module

In addition, we also need to provide complementary tsconfig files for the runtime/server and runtime/nitro folders, that extends `tsconfig.server.json` instead of `tsconfig.json`.
This is directly copied from the reference template for Nuxt3 apps at https://github.com/nuxt/starter/tree/v3

We further add a specific `.nuxtrc` file in the playground that allows autoImports in this folder only. This allows us to use classic Nuxt syntax (auto-imports) in the playground. This setup is replicated in fixture files for tests to run without modification.

Finally, the content of package.json is also upgraded to follow the Nuxt recommendations, specifically with regards to the `npm run dev:prepare` script, and to npm package versions.

## Checklist:

No changes are required to the documentation and the tests pass.

In order for the types to be correctly re-generated, it is required to run 
```
npm upgrade
npm run dev:prepare
```


vejja:
step 1: upgrade TS config

vejja:
step 2 : clean-up @ts-ignore overrides and 'h3' imports

vejja:
step 3 : remove 'h3' imports from playground and fixtures

vejja:
step 4 : use standard approach in nitro plugins

vejja:
step 5 : type runtimeConfig via nuxt/schema interface
Allows proper type inference for security options, avoids unecessary type re-casts

vejja:
step 6 : this one improve how the lru-cache storage is mounted
Currently when build we have the following warning:
```
Build is done with some warnings:
- Inlined implicit external unstorage
- Inlined implicit external destr
```
Instead of inlining the config and mounting later in middleware, we explicitly mount it upfront via config in the nitro:config hook
This allows to later invoke the standard `useStorage` utility in the rateLimiter middleware

Baroshem:
Hey,

I will make sure to merge it for 1.0.0-rc.2 :)

Baroshem:
@vejja 

I have included your latest fix for inlined implicit as it was blocking me from release. Thanks for that!

I think after the release of 1.0.0-rc.1 you could just revert the last commit and we will be able to merge the rest of your stuff :)

Baroshem:
Update.

Let's make it ready for the 1.0.0-rc.3. I needed to push a fix to 1.0.0-rc.1 because there was a bug that I didn't catch during testing. It was triggered once released to NPM :(

vejja:
> Update.
> 
> Let's make it ready for the 1.0.0-rc.3. I needed to push a fix to 1.0.0-rc.1 because there was a bug that I didn't catch during testing. It was triggered once released to NPM :(

Was the bug related to this PR? What is NPM complaining about?

Baroshem:
No no! the bug was releated to Nuxt internal.

Basically, when I used the await import to import the vite plugin it was working perfectly until publishing a package to npm.

After publishing and trying to install the package, it was breaking the app and showing error that the plugin is not a function. 

I changed the import from await to normal and it started working correctly. Pretty strange

package-lock.json
Baroshem, line 1: issue: The module uses `yarn` as the package manager. Could you please revert this change?

package.json
```diff
@@ -36,21 +36,21 @@
   ],
   "scripts": {
     "prepack": "nuxt-module-build",
-    "dev": "nuxt-module-build --stub && nuxi prepare playground && nuxi dev playground",
+    "dev": "nuxi dev playground",
```
Baroshem, line 39: suggestion: I have specifically added the module build in the dev stop to avoid situation where you would run dev without first building your module. I don't think we should be removing this.. Maybe we could just call here `dev:prepare` instead.

package.json
```diff
@@ -59,15 +59,15 @@
...
-    "nuxt": "^3.7.3",
-    "typescript": "5.2.2",
-    "vitest": "^0.28.5"
+    "@nuxt/module-builder": "^0.5.2",
+    "@nuxt/schema": "^3.7.4",
+    "@nuxt/test-utils": "^3.7.4",
+    "@nuxt/eslint-config": "^0.2.0",
+    "@types/node": "^18.18.1",
+    "eslint": "^8.50.0",
+    "nuxt": "^3.7.4",
```
Baroshem, line 68: suggestion: Nuxt recently released 3.8 version. Maybe we could migrate to that?

Baroshem (changes requested):
Few suggestions from my side

Baroshem:
@vejja could you please also resolve the conflict in 99-cspNonce file? It is probably caused by the other PR you created recently.

package-lock.json
vejja: ok

package.json
vejja: ok

package.json
vejja: ok

vejja:
> @vejja could you please also resolve the conflict in 99-cspNonce file? It is probably caused by the other PR you created recently.

will do

package.json
vejja: 3.8 requires modification to type imports
See https://github.com/nuxt/nuxt/releases/tag/v3.8.0
Only minor modifications are required in our codebase, I will include these as well

Baroshem:
Hey @vejja 

Could you please adjust this PR? I have merged your previous one about fixing complaine of CSP and it resulted in conflicts here.

Also, could you try to fix the CI?

Here is the output of Vercel:

image

Thanks! 💚

vejja:
Sure - Would you mind sharing your Vercel config file so that I can reproduce on my end ?

vejja:
Hi @Baroshem 
Do you know when the Vercel regression happened in the first time ? (the `Load plugin failed: pinceau/volar Error [ERR_REQUIRE_ESM]: require() of ES Module`)
I have it on my end on the main branch, and I'm trying to revert to a point where it did not exist

vejja:
Quick update:
- Merge conflicts resolved
- Vercel build now passing

Build-time warnings in the vercel logs are not on our side, but on third-party packages (nuxt-themes/docus).
Unfortunately there is nothing we can do. However, aside from polluting the logs, they do not prevent building successfully so it's a minor annoyance.

Baroshem (approved):

Baroshem:
Thanks @vejja ! I will merge it to the 1.0.0-rc.3 branch and test it there.

PR #256, opened 2023-10-20
Fix: make nonce implementation strict CSP compliant

trijpstra-fourlights:
## Types of changes

## Description
As reported by  @vejja in https://github.com/Baroshem/nuxt-security/discussions/241#discussioncomment-7326096 the  current nonce implementation is insecure by design.

While I was initially unconvinced, after going though all rendering modes and trying out various combinations I'm convinced that @vejja 's analysis is right and both the `nonce mode override` and `useNonce` composable should be removed.

My apologies to @vejja for not exploring your findings more thorough. You were completely in the right.

## Checklist:

### DIFF
--- a/src/runtime/composables/nonce.ts
+++ b/src/runtime/composables/nonce.ts
@@ -1,5 +1,0 @@
-import { useNuxtApp, useCookie } from '#imports'
-
-export function useNonce () {
-  return useNuxtApp().ssrContext?.event?.context.nonce ?? useCookie('nonce').value
-}
--- a/src/runtime/server/middleware/cspNonceHandler.ts
+++ b/src/runtime/server/middleware/cspNonceHandler.ts
@@ -5,7 +5,6 @@ import { getRouteRules } from '#imports'
 
 export type NonceOptions = {
   enabled: boolean;
-  mode: 'renew' | 'check';
   value: undefined | (() => string);
 }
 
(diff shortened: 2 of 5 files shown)


src/runtime/composables/nonce.ts
vejja, line 1: To be fair, you can keep the nonce composable server-side.
It could be useful to expose the nonce to third-party modules when they run server-side, such as NuxtImg as you rightly pointed out.
```
return useNuxtApp().ssrContext?.event?.context.nonce
```
and drop the useCookie part

src/runtime/server/middleware/cspNonceHandler.ts
```diff
@@ -5,7 +5,6 @@ import { getRouteRules } from '#imports'
 
 export type NonceOptions = {
   enabled: boolean;
-  mode: 'renew' | 'check';
   value: undefined | (() => string);
```
vejja, line 8: I'm not sure about giving the developper a way to override crypto primitives which are vetted to be safe

src/runtime/server/middleware/cspNonceHandler.ts
```diff
@@ -16,26 +15,8 @@ export default defineEventHandler((event) => {
...
-      }
-      case 'renew':
-      default: {
-        nonce = nonceConfig?.value ? nonceConfig.value() : Buffer.from(crypto.randomUUID()).toString('base64')
-        setCookie(event, 'nonce', nonce, { sameSite: true, secure: true })
-        event.context.nonce = nonce
-        break
-      }
-    }
+    const nonce = nonceConfig?.value ? nonceConfig.value() : Buffer.from(crypto.randomUUID()).toString('base64')
```
vejja, line 18: I'd leave it to just the crypto primitive

src/runtime/composables/nonce.ts
trijpstra-fourlights: IMO, this would then be the best solution: https://github.com/trijpstra-fourlights/nuxt-security/pull/1

as it allows the useNonce, even on SPA mode (although it will use the cookie in that case).

Interestingly enough, the  element does not have a nonce attribute according to mdn. So it doesn't apply there.

src/runtime/server/middleware/cspNonceHandler.ts
trijpstra-fourlights: IMO that's out of scope of this PR

trijpstra-fourlights:
I'm done with this. Please remove me as a contributor from this project.

PR #257, opened 2023-10-20
improve CSP compliance

vejja:
## Types of changes

## Description

These changes disable the possibility for the server to use a pre-determined nonce.

With these changes, the module now adheres to the W3C specs on CSP para 7.1: Nonce Reuse (https://www.w3.org/TR/CSP3/#security-nonces).

The modifications to the code include:
- disallowing the ability to set nonces on the server via cookie
- disallowing the possiblity to set nonces other than through cryptographically secure methods
- suppressing the former 'check' mode
- not sending nonces via cookie to the frontend

## Checklist:

### DIFF
--- a/docs/content/1.documentation/1.getting-started/2.configuration.md
+++ b/docs/content/1.documentation/1.getting-started/2.configuration.md
@@ -24,7 +24,7 @@ interface ModuleOptions {
   basicAuth: BasicAuth | false;
   enabled: boolean;
   csrf: CsrfOptions | false;
-  nonce: NonceOptions | false;
+  nonce: boolean;
   removeLoggers?: RemoveOptions | false;
   ssg?: Ssg;
 }
--- a/docs/content/1.documentation/2.headers/1.csp.md
+++ b/docs/content/1.documentation/2.headers/1.csp.md
@@ -193,11 +193,11 @@ The `nonce` value is generated per request and is added to the CSP header. This 
 ```ts
 export default defineNuxtConfig({
   routeRules: {
-    '/api/custom-route': {
-      nonce: false    // do not check nonce for this route (1)
+    '/custom-route': {
+      nonce: false    // do not generate nonce for this route (1)
     },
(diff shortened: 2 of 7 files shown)


Baroshem:
Hey @vejja 

Could you please also update the documentation of the module to be up to date with this change/fix? :)

vejja:
Hi @Baroshem 
I'm pushing a very minor change, otherwise the doc seems to be fine by me.

The only thing I'm unsure about, is our example about how to use nonce with `<NuxtImg>`. 
According to my tests, if we use this feature, we will have
- a `nonce` attribute in the `` element. As trijpstra-fourlights noted correctly, `` is *not nonceable* so this is useless
- two *identical* `nonce` attributes in the `<link preload>` element (if the `preload` option is used on `<NuxtImg>`). The first one is inserted by NuxtImg, and the second one by ourselves. So it is not required there either.

I can reach out to the people at NuxtImg to understand why they introduced this option in the first place if you think it is useful, but for now I think we can leave the section unchanged in our docs, as the above situation does not create any practical issue. 
In the longer term I suspect we could delete the section altogether, as users of nuxt-security probably don't need to use the `nonce` option when using `<NuxtImg>`.

vejja:
Background for this last commit:

CI seemingly fails at random times on the nonce test.
The issue arises in tests because when we extract the value of the nonce, it can sometimes contain the characters `+` or `/`, which are special regex characters, that therefore need to be backslashed escaped in the regex match.
The escape function is taken from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping

Note: this regression was introduced by the use of `crypto.randomBytes(16)` in [ee366c7](https://github.com/Baroshem/nuxt-security/pull/257/commits/ee366c72ff4be79def1b1dfac981b68538b00de0) in place of `crypto.randomUUID()`, which only generated alpha-numerical characters.

Baroshem (approved):

PR #267, opened 2023-10-26
Fix Basic Auth Configuration for Multiple Paths

rathahin:
## Types of changes

## Description

The existing implementation of basic authentication works partially when multiple routes are specified. The `.some()` method used in the code causes it to return as soon as one matching path is found, leading to incorrect behavior for additional paths. 

This change replaces the `.some()` method with individual variables to track the `exclude` and `include` lists. It uses `.some()` to update these variables and then checks them to decide whether to block the request.

This fix ensures that all specified paths in both the `exclude` and `include` arrays are correctly checked. It provides a clearer and more accurate way to handle basic authentication for multiple routes.

Resolves: #[Issue Number]

## Checklist:


pnpm-lock.yaml
Baroshem, line 1: issue: this is not needed for sure ;)

Baroshem (changes requested):
One small issue. If you revert it, it should be ready to review and further merge :)

pnpm-lock.yaml
rathahin: Thank you for pointing out the issue with `pnpm-lock.yaml`. I understand it's not needed and I'll revert it back to its original state. Once that's done, it should be ready for your review and potential merge.

Baroshem (approved):
Thanks!

Baroshem:
Hey @rathahin 

Could you please adjust the unit tests to fix the CI?

Baroshem:
Thanks @rathahin 💚

I will merge it now to the rc branch, conduct some tests and will release a new RC version today :)

Baroshem:
Thank you @rathahin ! 💚

PR #272, opened 2023-10-29
feat(csp): Extend CSP support of SSG mode

vejja:
## Types of changes

Extend CSP support of SSG applications:
- parse all HTML sections
- hashes for inline styles
- hashes for external scripts
- CSP headers for static nitro presets

## Description

This PR extends how we support CSP for SSG applications.

Currently, in SSG mode, Nuxt Security only searches for inline scripts in the `bodyAppend` section of Ntro's html rendering step.

With this PR, the following functionalities are added:
- Scan all sections of the `html` being rendered by Nitro. 
This covers the case where scripts or styles are added to the `<head>` tag, or to other parts of the `<body>` section.

- Scan inline styles, and add SHA hashes to the `style-src` policy. 
This covers the case where, in the absence of the `unsafe-inline` value, CSP requires that inline styles are individually authorized.

- Scan external script ressources that carry the `integrity` attribute and add SHA hashes to the `script-src` policy. 
This covers the case where an external ressource is not explicitely whitelisted (or where whitelisting is ignored by the use of `strict-dynamic`). 
In that case, CSP provides the ability to authorize an external script by its hash, provided this hash is present in its `integrity` attribute.

In addition, this PR improves SSG support for hosting providers that can automatically generate headers for static serving via their `nitro: preset` option. 
- For such hosting providers, the static server now returns HTTP headers with the correct CSP policy
- In all cases, the same policy is still being provided via the `<meta http-equiv>` tag. This ensures that, for hosting providers that cannot set HTTP headers, the static CSP policy is still being provided.

*Note: This PR does not make any change to CSP in SSR mode*

## Checklist:

### DIFF
--- a/.stackblitz/.gitignore
+++ b/.stackblitz/.gitignore
@@ -6,3 +6,4 @@ node_modules
 .output
 .env
 dist
+.vercel
--- a/src/module.ts
+++ b/src/module.ts
@@ -213,7 +213,7 @@ const removeCspHeaderForPrerenderedRoutes = (nuxt: Nuxt) => {
   const nitroRouteRules = nuxt.options.nitro.routeRules
   for (const route in nitroRouteRules) {
     const routeRules = nitroRouteRules[route]
-    if (routeRules.prerender) {
+    if (routeRules.prerender || nuxt.options.nitro.static) {
       routeRules.headers = routeRules.headers || {}
       routeRules.headers['Content-Security-Policy'] = ''
     }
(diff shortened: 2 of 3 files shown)


Baroshem:
Hey @vejja 

Could you please rebase your branch? 

Unless you think that it will be better to release this functionality in rc.4?

vejja:
> Hey @vejja
> 
> Could you please rebase your branch?
> 
> Unless you think that it will be better to release this functionality in rc.4?

Sure, done now

Baroshem:
Thanks @vejja really nice addition! :)

Baroshem (approved):

PR #284, opened 2023-11-07
fix(csrf): replace CSRF option `false` with `boolean`

Mohamed-Kaizen:
## Types of changes

## Description
image

**IDE**: webstorm

## Checklist:

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -9,26 +9,26 @@ export type Ssg = {
 };
 
 export interface ModuleOptions {
-  headers: SecurityHeaders | false;
-  requestSizeLimiter: RequestSizeLimiter | false;
-  rateLimiter: RateLimiter | false;
-  xssValidator: XssValidator | false;
-  corsHandler: CorsOptions | false;
-  allowedMethodsRestricter: AllowedHTTPMethods | false;
+  headers: SecurityHeaders | boolean;
+  requestSizeLimiter: RequestSizeLimiter | boolean;
+  rateLimiter: RateLimiter | boolean;
+  xssValidator: XssValidator | boolean;
+  corsHandler: CorsOptions | boolean;
+  allowedMethodsRestricter: AllowedHTTPMethods | boolean;
   hidePoweredBy: boolean;
-  basicAuth: BasicAuth | false;
+  basicAuth: BasicAuth | boolean;
   enabled: boolean;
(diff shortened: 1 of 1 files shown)


Baroshem:
Hey Buddy,

Thanks for creating a PR but I think that it is not correct. 

For CSRF, you should either set the `csrf` property in the config to an object (https://github.com/Morgbn/nuxt-csurf?tab=readme-ov-file#usage) or false to disable it if you dont need it. 

Same for other middlewares - Usually it is either about having them enabled with some configuration or disabled with `false`.

Only below ones can have true/false values:

```ts
  hidePoweredBy: boolean;
  enabled: boolean;
  nonce: boolean;
```

So the current behavior is correct and intentional.

I am glad for your contribution though so I would like to recommend you for the future to open an issue/question first. Then it will allow to avoid work that cannot be merged later on :)

Mohamed-Kaizen:
> For CSRF, you should either set the csrf property in the config to an object (https://github.com/Morgbn/nuxt-csurf?tab=readme-ov-file#usage) or false to disable it if you dont need it.

The [documentation](https://nuxt-security.vercel.app/documentation/middleware/csrf)  says that you can enable CSRF by setting `security.csrf` to `true` in your Nuxt configuration.However, this does not work with TypeScript. I believe that either the TypeScript types or the documentation need to be updated to reflect this.

Additionally, I believe that having `security.csrf`  set to  `true`  would make it easier for users who want the recommended configuration, unless they need to customize it. The same is true for other middlewares.

>I am glad for your contribution though so I would like to recommend you for the future to open an issue/question first. Then it will allow to avoid work that cannot be merged later on :)

Yeah,  i will do that

Baroshem:
Hey @Mohamed-Kaizen 

indeed you are correct. The docs says it correctly while the type here is not correct.

Instead of:

```ts
export interface ModuleOptions {
  csrf: CsrfOptions | false;
}
```

It should be:

```ts
export interface ModuleOptions {
  csrf: CsrfOptions | boolean;
}
```

Could you please correct just this line? Others are correct :)

vejja:
@Baroshem Just fixed this one, can be merged

Baroshem:
Awesome, thank you @vejja 

I will release it with the upcoming 1.0.0 next week as it is a really small bug fix ;)

Baroshem (approved):

PR #289, opened 2023-11-14
feat(csp): Strict CSP by default

### LINKED ISSUE #279: fix: hash generation missing for link as script
### Steps to reproduce

Add a `link` header with `as="script"` attribute `href`ing to a script.

### What is Expected?

There should be a hash added to the `script-src` for that linked script.

I think only `script` tags are accounted for currently and thus `<link as="script"` tags are not whitelisted by having a hash added to `script-src`.

Stylesheets could benefit from that as well, I guess.

vejja:
## Types of changes

This PR modifies the default configuration of Nuxt Security to enable Strict CSP by default.

Closes #279

## Checklist:


src/defaultConfig.ts
```diff
@@ -8,14 +8,15 @@ export const defaultSecurityConfig = (serverlUrl: string): ModuleOptions => ({
...
-      'base-uri': ["'self'"],
+      'base-uri': ["'none'"],
       'font-src': ["'self'", 'https:', 'data:'],
       'form-action': ["'self'"],
       'frame-ancestors': ["'self'"],
       'img-src': ["'self'", 'data:'],
       'object-src': ["'none'"],
       'script-src-attr': ["'none'"],
       'style-src': ["'self'", 'https:', "'unsafe-inline'"],
+      'script-src': ["'self'", 'https:', "'unsafe-inline'", "'strict-dynamic'", "'nonce-{{nonce}}'"],
```
Baroshem, line 19: issue: I tested the app with this setup and I get the following error. Could you please check it out?

image

Baroshem (changes requested):

src/defaultConfig.ts
vejja: Looks like the random nonce value is not set in the headers. They still have the placeholders.
Do you have `{ nonce: true }` ? It should be the default

Baroshem: Ah interesting. I had a duplicate property for nonce (both true an false but in different places in the config) but TS didn't catch it.

I removed the duplicate one and it works correctly.

Thanks for checking that. Works like a charm. I will merge it right now :)

Baroshem (approved):

src/defaultConfig.ts
vejja: Let me double check though, I believe you might have caught an edge case in dev mode

vejja: @Baroshem Can you let me know what was your double config setup ?
I still find it weird that the random nonce was not in the headers

Issue #288, opened 2023-11-13
Strict-Transport-Security as string not parsing max-age correctly.


KyleBrown-804:
### Version
nuxt-security: v0.14.4
nuxt: v3.7.1

### Reproduction Link
Not able to supply this.

### Steps to reproduce
1. Implement Strict-Transport-Security header as a string per the accepted type def here: https://github.com/Baroshem/nuxt-security/blob/main/src/types/headers.ts#L205.
Example:
```
security: {
    ...,
    headers: {
        strictTransportSecurity: 'max-age=15552000; includeSubDomains'
    }
}
```

### What is Expected?
The response headers should be in proper format as 'max-age=<value here>; includeSubDomains'.

### What is actually happening?
max-age coming back as undefined, includeSubDomains not in header value either as shown in image below:
<img width="311" alt="Screenshot 2023-11-13 at 12 38 39 PM" src="https://github.com/Baroshem/nuxt-security/assets/56103801/7d884703-64ce-4633-bbe4-e746311eec09">

It is worth noting that opting for passing the object instead of a string as such fixes the issue, however string should be a valid option according to the type def:

Example:
```
security: {
...,
    headers: {
        strictTransportSecurity: {
            maxAge: 15552000,
            includeSubdomains: true
        }
    }
}
```
Will yield the following:
<img width="406" alt="Screenshot 2023-11-13 at 12 38 59 PM" src="https://github.com/Baroshem/nuxt-security/assets/56103801/34765199-46f2-495e-a542-123fdfc936bd">

Baroshem:
Hey,

Thanks for reporting this issue.

 I have tested it on 0.14.4 and I can confirm that it does not work there. However, I have tested it on the 1.0.0-rc.3 version and it is working even when passed as a string.

Would you be able to migrate to the RC version? I have included the Migration Guide in release notes so it should be easier to migrate. 

I am also planning to release a stable 1.0.0 version in the next few weeks if you are not sure about using the RC version :)

PR #296, opened 2023-11-16
docs: update route rules docs

Baroshem:
## Types of changes

## Checklist:

### DIFF
--- a/docs/content/1.documentation/2.headers/1.csp.md
+++ b/docs/content/1.documentation/2.headers/1.csp.md
@@ -29,7 +29,7 @@ export default defineNuxtConfig({
   routeRules: {
     '/custom-route': {
       headers: {
-        'Content-Security-Policy': <OPTIONS>
+        'Content-Security-Policy': 'default-src 'self'; img-src https://*; child-src 'none';'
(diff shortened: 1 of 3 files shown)


Baroshem:
@vejja please check if this makes more sense.

vejja:
> @vejja please check if this makes more sense.

Absolutely, it does - small details only
By the way #288 is related (OP is setting `strictTransportSecurity`on the custom security property instead of using the native radix header)

docs/content/1.documentation/2.headers/1.csp.md
```diff
@@ -29,7 +29,7 @@ export default defineNuxtConfig({
   routeRules: {
     '/custom-route': {
       headers: {
-        'Content-Security-Policy': <OPTIONS>
+        'Content-Security-Policy': 'default-src 'self'; img-src https://*; child-src 'none';'
```
vejja, line 32: "default-src 'self'; img-src https:; child-src 'none';"

Baroshem: Thanks, fixed!

PR #298, opened 2023-11-19
feat: allow configuring headers in runtime

### LINKED ISSUE #258: feat: runtime header configuration

I'd like to be able to set content-security-policy headers depending on the `host`. For example:

```ts
'connect-src': [
    `https://strapi.${host}`,
],
```

I'd like to be able to dynamically specify security headers per request. I want to be able to define security headers using a function that has a nitro request `event` as parameter.

Before using this module I used a middleware and `event.node.req.headers.host` to extract the `host`.

### Additional context

I think https://github.com/Baroshem/nuxt-security/pull/233 goes into the right direction. But I think that PR only makes the definition at runtime instead of at build time while still making use of the static module configuration.

huang-julien:
resolve #258 

## Types of changes

## Description

Hi :wave: 

This PR allows to configure headers configuration in runtime. 

users will simply have to call `nuxt-security:headers` hook to change the configuration. The new configuration will have a higher priority than the routeRules .

Strangely, auto scanned plugins have a higher priority than registered plugins.  So I also created the `nuxt-security:ready` nitro hook. 

## Checklist:


huang-julien:
@dargmuesli @Baroshem @harlan-zw Would love to have your feedback !

 @pi0 feel  free to correct this PR is it is not correct 😅 , i really hope this looks good to you

dargmuesli (commented):
Looks good at the first glance! I will try it out later :heart:

Baroshem:
Hey, amazing work!

Would you be ok if we could merge it to the 1.1.0 version that I am planning to release in December?

This week, I would love to release 1.0.0 and your change would work perfectly for the first minor update after stable 1.0.0 :)

huang-julien:
Sure ! I guess we can also iterate on `00-context` to allow more configurable options in the future but users are mostly asking for configurable CSP (including myself) for now

src/types/index.ts
```diff
@@ -23,6 +22,10 @@ export interface ModuleOptions {
   nonce: boolean;
   removeLoggers?: RemoveOptions | false;
   ssg?: Ssg;
+  /**
+   * enable runtime nitro hooks to configure some options at runtime
+   */
+  runtimeHooks: boolean;
```
harlan-zw, line 28: I think you'd be better off omitting this as it may get confusing and buggy if other runtime hooks are added in the future for no perceived value (that I can think of).

src/types/index.ts
```diff
@@ -34,3 +37,20 @@ export interface NuxtSecurityRouteRules {
...
+
+declare module 'h3' {
+  interface H3EventContext {
+      security: NuxtSecurityEventContext
+  }
+}
+
+declare module 'nitropack' {
+  interface NitroRuntimeHooks {
+    'nuxt-security:headers': (route: string, headers: SecurityHeaders) => void
```
harlan-zw, line 53: I haven't dug into how the module works but this type of augmentation generally doesn't work when users install the module.

You need to use the runtime-generated types for this sort of thing usually.

Also I'd type it as `Promise<void> | void` so users can use promises.

harlan-zw: Personally I'd use a single context object instead of two separate params and I'd use a H3Event over the `route`.

Not required though

src/runtime/nitro/plugins/00-context.ts
```diff
@@ -0,0 +1,24 @@
...
+
+        for (const [header, headerOptions] of Object.entries(headersConfig)) {
+            headers[SECURITY_HEADER_NAMES[header]] = headerOptions === false ? false : getHeaderValueFromOptions(header as HeaderMapper, headerOptions as any)
+        }
+
+        router.insert(route, headers)
+    })
+
+    nitroApp.hooks.hook('request', (event) => {
+        event.context.security = event.context.security || {}
```
harlan-zw, line 19: I'd use the module name to avoid conflicts

src/runtime/nitro/plugins/00-context.ts
```diff
@@ -0,0 +1,24 @@
+import { type HeaderMapper, getHeaderValueFromOptions, SECURITY_HEADER_NAMES } from "../../utils/headers"
+import { createRouter} from "radix3"
+import { defineNitroPlugin } from '#imports'
+
+export default defineNitroPlugin((nitroApp) => {
+    const router = createRouter()
+
+    nitroApp.hooks.hook('nuxt-security:headers', (route, headersConfig) => {
```
harlan-zw, line 8: I have a feeling this would be more useful if it was called on each request, otherwise the user should just use route rules?

src/types/index.ts
```diff
@@ -34,3 +37,20 @@ export interface NuxtSecurityRouteRules {
   allowedMethodsRestricter?: AllowedHTTPMethods | false;
   nonce?: boolean;
 }
+
+export interface NuxtSecurityEventContext {
+  headers: Record<string, string | string[] | number | false> | null
```
harlan-zw, line 42: Should this share the types of `SecurityHeaders`?

src/runtime/nitro/plugins/00-context.ts
huang-julien: i used security because it's the module `configKey`, but yeah we can change it to `nuxtSecurity` !

src/runtime/nitro/plugins/00-context.ts
huang-julien: Do you mean it would be up to us to call a hook insteading of hooking into the event ? 

I also thought of this. The reason I made nuxt-security hooking into it instead of calling the hook is because we would need to `await` all hooks. 

So if users are doing a long async hooks, this would block their response but it also makes sense !

src/types/index.ts
huang-julien: :+1:

src/types/index.ts
huang-julien: oops indeed

src/types/index.ts
huang-julien: yes, we probably need to add a reference in a `d.ts` file. I'll verify if this works later !

huang-julien: would you prefer something like this?

```ts
'nuxt-security:headers': ({
  route: string,
  headers: SecurityHeaders
}) => void
```

harlan-zw: yes, it just makes it easier if you want to add other things into the hook context later

src/runtime/nitro/plugins/00-context.ts
harlan-zw: I think if ends users will use this context object it could make sense with this name, otherwise if it's just internal module name is better.

src/runtime/nitro/plugins/00-context.ts
harlan-zw: I need to dig into the code a bit deeper to answer this, think I'm missing some context. Will get back to you when I have a chance (otherwise feel free to ping me)

huang-julien: :+1: feel free to ping/call me, maybe can we try to find a better implementation together with @dargmuesli and @Baroshem

src/runtime/nitro/plugins/00-context.ts
Baroshem: I would stick to `security` instead of `nuxt-security` purely because this is a config key for both nuxt.config.ts file and for route rules so it would be better IMHO to keep this consistent :)

Baroshem:
@huang-julien 

Sorry for waiting so much for my feedback on this great PR. I am back again and could take a look at it .

Do you need me to answer any particular discussion?

src/runtime/server/middleware/headers.ts
```diff
@@ -0,0 +1,13 @@
+import { defineEventHandler, setHeader, removeResponseHeader } from '#imports'
+
+export default defineEventHandler((event) => {
+    if(event.context.security.headers) {  
+        Object.entries(event.context.security.headers).forEach(([header, value]) => {
+            if(value === false) {
+                removeResponseHeader(event, header)
+            }else {
+                setHeader(event, header, value, )
```
Baroshem, line 9: question: is this correct? There is `value, ` at the end.

docs/content/1.documentation/2.headers/15.runtimeConfiguration.md
```diff
@@ -0,0 +1,37 @@
...
+    nitroApp.hooks.hook('nuxt-security:ready', () => {
+        nitroApp.hooks.callHook('nuxt-security:headers', '/**' ,{
+            contentSecurityPolicy: {
+                "script-src": ["'self'", "'unsafe-inline'"],
+            },
+            xFrameOptions: false
+        })
+    })
+})
+```
```
Baroshem, line 37: suggestion: I would recommend to move this documentation to `Usage` page (https://nuxt-security.vercel.app/documentation/getting-started/usage) to keep docs consistent

playground/nuxt.config.ts
```diff
@@ -24,6 +24,7 @@ export default defineNuxtConfig({
     rateLimiter: {
       tokensPerInterval: 10,
       interval: 10000
-    }
+    },
+    runtimeHooks: true
```
Baroshem, line 28: question: I wonder if this should actually be enabled by default. I think that majority of the users will just use the nuxt.config.ts config or route rules while this runtime configuration is a bit of an edge case (at least IMO).

With these kind of configuration, I usually mark them as optional and disabled by default (to not ship unused code to other users).

So, if you want to have a runtime configuration, just set this config value `security.runtimeHooks = false` and it will work for you.

WDYT?

Baroshem (commented):
Few questions/suggestions

src/runtime/server/middleware/headers.ts
huang-julien: oops, i need to lint that

playground/nuxt.config.ts
huang-julien: oh i didn't remove this since the removal of the `runtimeHooks` option

@harlan-zw think we don't really need this option to always have a runtime configuration of headers. So should we set back `runtimeHooks` in the options ?

If we set back `runtimeHooks` option, i think it should be default to false. Then if true, we'll add the nitro plugin

Baroshem: I personally would recommend to add an option and make it disabled by default to not add additional code for users that may not use it.

So add an option but make it disabled by default and make a note in the docs that in order to use this feature, set this config option to true and then, call the hook (like you have explained in the docs)

I think this delivers the functionality to the user that may need it without forcing every user of the module to have some dead code.

harlan-zw:
Having reviewed the source and looking at this PR again, I think the best solution would be to introduce a composable to resolve the security route rules.

Currently, each function is doing this:

```ts
const { security } = getRouteRules(event)
```

The PR to my understanding is solving the issue of users wanting to modify this `security` var at runtime. So to solve this, introducing a new composable that wraps this logic while also exposing a Nitro hook should solve the issue.

```ts
export async function resolveSecurityContext(e: H3Event) {
  const routeRules = getRouteRules(event).security
  const nitroApp = useNitroApp()
  const context = { rules: { ...routeRules } } // note: you may need to do a proper clone here (see klona)
  await nitroApp.hooks.hook('nuxt-security:context', context)
  return context
}
```

```ts
const security = await resolveSecurityContext(event)
```

The only catch is that you don't want users running slow events in this hook. So I'd update the docs with an example like so.

```ts
  const myCustomSecurityRules = await $fetch('/security')
  nitroApp.hooks.hook('nuxt-security:context', (context) => {
      context.rules = defu(context.rules, myCustomSecurityRules)
  })
})
```

Let me know what you think of this and if it solves the issue.

Baroshem:
@huang-julien any update from your side? I would love to plan 1.1.0 release with this feature :)

huang-julien:
Yup probably this weekend, i didn't had a lot of time since the end of holidays :sob:

huang-julien:
- docs moved to `usage` following @Baroshem's suggestion
- lint
- set back runtimeHooks option  following @Baroshem's suggestion
- fix: added type  augmentation
- transformed the hook parameters into a context object following @harlan-zw 's suggestion

> Having reviewed the source and looking at this PR again, I think the best solution would be to introduce a composable to resolve the security route rules.
> 
> Currently, each function is doing this:
> 
> ```ts
> // nuxt-security plugin
> const { security } = getRouteRules(event)
> ```
> 
> The PR to my understanding is solving the issue of users wanting to modify this `security` var at runtime. In theory, a user could just modify the route rules within a nitro plugin `event.context._nitro.routeRules`, but the DX isn't great.
> 
> Instead, using a new composable that wraps this logic while also exposing a Nitro hook should solve the issue.
> 
> ```ts
> export async function resolveSecurityContext(e: H3Event) {
>   const routeRules = getRouteRules(event).security
>   const nitroApp = useNitroApp()
>   const context = { rules: { ...routeRules } } // can't remember if this is frozen, you may need to do a clone (see klona)
>   await nitroApp.hooks.hook('nuxt-security:context', context)
>   return context
> }
> ```
> 
> ```ts
> // nuxt-security plugin
> const security = await resolveSecurityContext(event)
> ```
> 
> The only catch is that you don't want users running slow events in this hook. So I'd update the docs with an example like so.
> 
> ```ts
> export default defineNitroPlugin((nitroApp) => {
>   const myCustomSecurityRules = await $fetch('/security')
>   nitroApp.hooks.hook('nuxt-security:context', (context) => {
>       context.rules = defu(context.rules, myCustomSecurityRules)
>   })
> })
> ```
> 
> Let me know what you think of this and if it solves the issue. Also feel free to name the composable / hook to whatever you think makes the most sense.

I think this can probably have it's own PR. I completly agree that having a composable will be better in terms of DX :+1:  . The only issue i see is the hook being possibly async which can block the request, that's why it's up to the user to call the hook instead the other way in this PR.
What do you think @harlan-zw ?

harlan-zw:
Separate PR is good with me, whatever is easy. I think as long as it's clear to the user it's not an issue, that should be as simple as documentation and the fact that all server hooks are async and block the request.

vejja:
Hi @huang-julien 
I'd like to consider introducing a deeper refactoring of our Nitro plugins that would be based on your approach.
Your PR seems to provide a proper foundation for solving other issues, e.g. #335 which is fundamentally caused by us trying to resolve headers in various parts of the module.
Would you mind if we discuss this in another thread ?

Baroshem:
Looks good from my side now :)

huang-julien:
Hi @vejja 👋 
Sure we can :)

vejja:
> Hi @vejja 👋 Sure we can :)

Thanks !
Let's move it to https://github.com/Baroshem/nuxt-security/discussions/346
I am basically trying to lay out my understanding of how it works, would be extremely happy to have your feedback - especially on the assumptions I'm making about your PR

Baroshem (approved):

Baroshem:
@huang-julien @harlan-zw @vejja @dargmuesli 

I am merging this PR. Thanks for all the work! 💚

test/fixtures/runtime-hooks/nuxt.config.ts
```diff
@@ -0,0 +1,21 @@
...
+      headers: {
+        'x-xss-protection': '1',
+      }
+    }
+  },
+  security: {
+    nonce: false,
+    runtimeHooks: true,
+    headers: {
+      contentSecurityPolicy: false
```
dargmuesli, line 18: @huang-julien is it required to turn off nonces and content security policy headers here?

When I keep `contentSecurityPolicy` headers enabled here, the headers set in the hook are not applied it seems.
When I disable `contentSecurityPolicy` headers here, the headers set in the hook are applied but without `{{nonce}}` replacement. Is that intended?

vejja:
> Having reviewed the source and looking at this PR again, I think the best solution would be to introduce a composable to resolve the security route rules.
> 
> Currently, each function is doing this:
> 
> ```ts
> // nuxt-security plugin
> const { security } = getRouteRules(event)
> ```
> 
> The PR to my understanding is solving the issue of users wanting to modify this `security` var at runtime. In theory, a user could just modify the route rules within a nitro plugin `event.context._nitro.routeRules`, but the DX isn't great.
> 
> Instead, using a new composable that wraps this logic while also exposing a Nitro hook should solve the issue.
> 
> ```ts
> export async function resolveSecurityContext(e: H3Event) {
>   const routeRules = getRouteRules(event).security
>   const nitroApp = useNitroApp()
>   const context = { rules: { ...routeRules } } // can't remember if this is frozen, you may need to do a clone (see klona)
>   await nitroApp.hooks.hook('nuxt-security:context', context)
>   return context
> }
> ```
> 
>  ```ts
> // nuxt-security plugin
> const security = await resolveSecurityContext(event)
> ```
> 
> The only catch is that you don't want users running slow events in this hook. So I'd update the docs with an example like so.
> 
> ```ts
> export default defineNitroPlugin((nitroApp) => {
>   const myCustomSecurityRules = await $fetch('/security')
>   nitroApp.hooks.hook('nuxt-security:context', (context) => {
>       context.rules = defu(context.rules, myCustomSecurityRules)
>   })
> })
> ```
> 
> Let me know what you think of this and if it solves the issue. Also feel free to name the composable / hook to whatever you think makes the most sense.

@harlan-zw 
Currently trying to implement this
2 questions here
1. in `resolveSecurityContext` , 4th line should it be `callHook` instead of `hook` ? i.e.
  ```ts
  await nitroApp.hooks.callHook('nuxt-security:context', context)
  ```

2. in `defineNitroPlugin`, I can't seem to make it `async(nitroApp)`, which is required for `await $fetch()`. i.e.
  ```ts
  export default defineNitroPlugin(async(nitroApp) => {
     ...
  })
  ```
  Looks like `defineNitroPlugin` cannot be async
  Any idea on how to await fetching headers from within the Nitro plugin ?

kalnode:
Hello, apologies for gravedigging.

I've scrubbed through various discussions and think the work here applies to what I need (also [similar to this thread](https://github.com/nuxt/nuxt/discussions/24905)).

Ultimately I'm striving for:
1 - Configure everything through env variable (such that no hard URLs are in the codebase)
2 - Centralize all config (ie nuxt.config.ts), however I'd live with a nitro plugin, so as long as # 1 is satisfied.

I'm simply looking to pass env variables to `contentSecurityPolicy`. I understand there's challenges in doing this via `nuxt.config.ts`. I've spun my wheels for hours on this and just taking pot shots.

STARTING POINT
---

**nuxt.config.ts**
```
// STARTING POINT
// Works, but need the URL to come from env variables
export default defineNuxtConfig({
    security: {
            headers: {
                contentSecurityPolicy: {                
                    // DESIRED, BUT FAILS - Ideally, we pass env value
                    'img-src': ["'self'", 'data:', process.env.SOMEURL] 
                    
                    // WORKS, BUT NOT IDEAL
                    'img-src': ["'self'", 'data:', 'https://some.website.net']
            }
        }
    }
}
```

TRYING NITROPLUGIN INSTEAD
---

Based on discussions, it's been suggested to use a custom NitroPlugin and hooks to modify CSP elsewhere, and perhaps in this way be able to pass env variables. Below I'm trying out this snippet: https://nuxt-security.vercel.app/advanced/hooks#available-hooks. The plugin works (can change CSP per-route), however I'm still unable to pass env variables.

Is there something additional needed?

**server/myplugin/myplugin.ts**
```
    nitroApp.hooks.hook('nuxt-security:routeRules', async (appSecurityOptions) => {
        appSecurityOptions['/**'] = defuReplaceArray(
            {
                headers: {
                    contentSecurityPolicy: {                        

                        // Fails - getting undefined/null on these gets
                        //"img-src": ["'self'", 'data:', process.env.IMAGEBASEURL ]
                        //"img-src": ["'self'", 'data:', useRuntimeConfig().public.IMAGEBASEURL ]
                        
                        // Works, however still need to make this URL come from env variables
                        "img-src": ["'self'", 'data:', 'https://some.website.net' ]
                    }
                }
            },
            appSecurityOptions['/**']
        )
    })
})
```

Am I going in the wrong direction, or has there been another way since

dargmuesli:
@kalnode please open a new discussion or issue

PR #304, opened 2023-11-27
feat(chore): Headers per route

### LINKED ISSUE #288: Strict-Transport-Security as string not parsing max-age correctly.

### LINKED ISSUE #290: Migrate to Node 18.X

### LINKED ISSUE #292: Per-route object based headers configuration
It is not possible to define CSP with the `ContentSecurityPolicy` object type in the per-route configuration.

### Version

nuxt-security:  1.0.0-rc.4
nuxt:  3.8.0

### Reproduction Link

Stackblitz:
https://stackblitz.com/edit/nuxt-starter-pf1ssj?file=app.vue

### Steps to reproduce

Open Stackblitz on home page and observe console errors.

Object format is not supported and CSP header has an invalid value:

Capture d’écran 2023-11-15 à 16 00 39

In addition, TypeScript complains about the `object` type:

Capture d’écran 2023-11-15 à 16 09 02

### LINKED ISSUE #297: failed to find a valid digest in the 'integrity' attribute?
I recently updated to rc4 and all users got the following error when I deployed to production:

image

Is this related to the update? Is it a bug, or would there be a changelog somewhere that would indicate what changes I might need to make?

### LINKED ISSUE #305: Nuxt 3.8.1 breaks Subresource Integrity
No SRI attributes in SSG mode

### Version

nuxt-security:  1.0.0-rc.4
nuxt:  ^3.8.1

### Reproduction Link

https://stackblitz.com/edit/nuxt-starter-3x8r6t?file=README.md

### Steps to reproduce

Open Stackblitz repro with DevTools, inspect Response

### What is Expected?

`<head>` elements should contain `integrity` attributes

`<head>` elements do not contain `integrity` attributes

### Issue description

The issue only appears in SSG mode. SSR mode is not affected.
The Stackblitz repro installs in SSG mode to illustrate the issue.
Capture d’écran 2023-11-28 à 11 06 18

You can modify the `nuxt` version to 3.8.0 and the issue will disappear
Capture d’écran 2023-11-28 à 11 06 05

### Origin of issue

Nuxt 3.8.1 modified the behavior of the `nitro:build:public-assets` hook.
- Before 3.8.1, this hook was called before pre-rendering pages
- After 3.8.1, this hook is now called after pre-rendering pages

Responsible commit was https://github.com/nuxt/nuxt/commit/a4b333bca2a729b2629d45212b12eb8e046f47be#diff-85b4fe47ac4284e88552626029d5147793563b4dcdab205cd8dc6afb190222aaR462

### Resolution

PR #304 will fix

### LINKED ISSUE #307: Unrecognized Content-Security-Policy directive 'undefined'.
### Version

nuxt-security: 1.0.0-rc.4
nuxt: 3.7.4

### Steps to reproduce

nuxt.config.ts

```ts
export default defineNuxtConfig({
  ...
  modules: ['nuxt-security'],
  security: {
    headers: {
      crossOriginOpenerPolicy: false,
      crossOriginEmbedderPolicy: false,
      contentSecurityPolicy: false,
    },
  },
...
});
```

### What is Expected?

This warning shouldn't be printed in the browser console because `contentSecurityPolicy` is set to false.

Got: `Unrecognized Content-Security-Policy directive 'undefined'.` in browser console.

vejja:
### Enables headers configuration on a per-route level
 
This PR introduces per-route configuration of security headers, via
```ts
defineNuxtConfig({
  routeRules: {
    [some-route]: {
      security: {
        headers : ...
      }
    }
  }
})
```

Closes #292 
Closes #291 
Closes #297 
Closes #288 
Closes #146 
Closes #305 
Closes #290 
Closes #307 

## Types of changes

This PR is a non-breaking change

### Current Problem

Currently, Nuxt Security supports defining headers **and** middlewares on a per-route basis, by implementing a custom `security` property in the Nitro `routeRules`. This allows the module to resolve per-route security options on the fly.

However, several limitations prevent the correct definition of per-route **headers** :

1. While per-route middleware must be defined within the `security` custom property, per-route headers must be defined within the standard `headers` property of the Nitro `routeRules`;
2. The standard `headers` property uses the HTTP header names (e.g. `'Permissions-Policy'`), while the security headers property uses custom names (e.g. `permissionsPolicy`), as defined in type `SecurityHeaders`;
3. The standard `headers` property uses HTTP header string values (e.g. `'geolocation=(), camera=()'`), while the security headers uses custom object syntax (e.g. `{geolocation: [], camera: []}`), also defined in type `SecurityHeaders`.

As a consequence, our users get misled by the TypeScript linter, which tries to enforce string syntax on the standard `headers` property where we are in fact requiring object syntax for proper parsing. 

### Proposed Solution

With this PR, we move the definition of SecurityHeaders under the `security` property. They are now located in the same place as middlewares, and are not defined via the standard `headers` property anymore.

This provides our users with a consistent way of defining per-route security options, together with correct TypeScript definitions for the object syntax requirement.

### Compatibility Considerations

This PR is built with full backwards-compatibility as a pre-requisite: 
- Legacy code that departed from the TypeScript linter recommendation by using object syntax for the standard `headers` property MUST continue to run: the module now scans standard headers and copies relevant `SecurityHeaders` internally to the correct place
- Legacy code that followed the TypeScript linter recommendations by using string syntax under the `headers` property MUST continue to run: the module also detects whether the previous scan finds strings and parses them into object syntax.

### Conflict Considerations

With this PR, headers can now be defined in two places, which implies defining a conflict resolution rule:
- If a security header was defined both under the standard `headers` property and under the Nuxt Security module `security` custom property, Nuxt Security has priority
- All other standard headers MUST continue to apply, irrespective of Nuxt Security

Internally we achieve this by merging the headers into the `security` property, rather than merging them into the `headers` property as was the case previously. We keep a clean, unmodified definition of the standard `headers` all along the `routeRules` tree, which allows our users to continue using the native Nuxt `routeRules` property. This is particularly useful if they want to define specific headers for their `/api` routes or for their static assets.

### Nesting Considerations

This PR fully supports recursive resolution of routes. For instance, if a rule is defined for `/route/**` and another one is defined for `/route/page`, the module MUST resolve the routes recursively.

Internally we do not implement a custom algorithm for this. We simply delegate nested route resolution to the native Nitro `getRouteRules` function, which works seamlessly for the custom `security` property. As a result, nested headers definitions are now resolved identically to other `routeRules`.

Because under the hood, Nitro uses `defu` to recursively merge route rules, this allows us to introduce several additional features:
 
- It is now possible to remove a header by setting it to `false`. Previously this was not possible, the only workaround was to set it to `''` empty string. For instance, CSP can now be disabled with `contentSecurityPolicy : false`;

- For `SecurityHeaders` using the object syntax (namely: `permissionsPolicy`, `contentSecurityPolicy` and `strictTransportSecurity`), nested properties can now be individually introduced or deactivated recursively. 
  For instance, if a higher route has `strictTransportSecurity: { maxAge: 1000, includeSubdomains: true, preload: true }` and a nested route has `strictTransportSecurity: { maxAge: 100, preload: false }`, the nested route resolves naturally to `'Strict-Transport-Security': 'max-age=100; includeSubDomains;'`;

- For `SecurityHeaders` whose individual properties use the array syntax (namely : `permissionsPolicy` and `contentSecurityPolicy`), it is now possible to decide whether the properties will be merged additively or substitutively.
  This is being made possible thanks to `defu`, which merges arrays additively (by concatenation) but merges strings substitutively (by replacement). We now give the ability to use string syntax instead of array syntax to make use of this possibility.

### Other features

This PR also fixes a number of unrelated issues:
- Node 16 deprecated as of Nuxt 3.8.2 (#290): CI is upgraded to Node 18
- Subresource Integrity broken by Nuxt 3.8.1 (#305) : SRI hashes for SSG restored via earlier pre-rendering hook
- Security Headers for HTML-only resources (#146) : Headers are now generated from `render:html` hook
- Post-Build Processing invalidates SRI (#297) : Via new documentation section

## Checklist:

All the considerations laid out above are extensively tested under a the new `perRoute` suite which includes 31 new tests.


vejja:
> Hey @vejja
> 
> Do you think it would be possible to extract only bugfixes from this PR and make from them an RC.5 release just before the stable 1.0.0 and later for 1.1.0 release the support for security per route?
> 
> Or should we merge this PR now with all the commits for rc.5 and then later release a stable 1.0.0 and for 1.1.0 plan pther features?

Hi @Baroshem 
We can merge everything in rc.5 I think. Extracting only bug fixes will be a bit difficult but doable, at least partially
But it's up to you, let me know

Baroshem:
@vejja 

Lets create then a new rc version with these changes after I will be back.

And after that, lets aim to release a stable 1.0.0 next in December :)

Baroshem:
Hey @vejja 

I will merge this PR to rc.5 branch and start testing it so that we could release rc.5 this week :)

Issue #335, opened 2023-12-28
Error deploying to Cloudflare Workers after upgrade beyond rc.2


nicgene:
### Version

- nuxt-security: 1.0.0
- Node.js 18.18.0
- Nuxt 3.9.0
- Nitro 2.8.1

### Steps to reproduce

Upgrading nuxt-security beyond 1.0.0-rc.2 breaks deployment to Cloudflare Workers using preset cloudflare_module.

`ERROR  Cannot resolve "node-fetch-native/native/proxy" from "..." and externals are not allowed!`

Baroshem:
Hey Buddy,

Thanks for reporting this issue.

Could you share your nuxt.config.ts file?

Current issue details does not ring a bell for me :(

nicgene:
```
export default defineNuxtConfig({
  modules: [
    "nuxt-security",
  ],
  security: {
    headers: {
      crossOriginEmbedderPolicy: process.env.NODE_ENV === "development" ? "unsafe-none" : "require-corp",
      contentSecurityPolicy: {
        "img-src": [
          "'self'",
          "data:",
          "https://imagedelivery.net"
        ],
        "script-src": [
          "'nonce-{{nonce}}'",
          "'strict-dynamic'",
          "'unsafe-eval'"
        ]
      }
    },
    rateLimiter: false
  }
});

```

nicgene:
https://github.com/nuxt/image/issues/1168

Baroshem:
Thanks for sharing the nuxt configuration, the file look fine.

It does not look to me like there is some kind of issue with NuxtSecurity.

Could you try disabling the module or removing it for a test to see if it will still result in a bug after deploying to Cloudflare?

nicgene:
Deployment to Cloudflare works with nuxt-security disabled. I have this issue on two Nuxt sites deployed to Workers. I just tried updating the other site from rc.2 to 1.0.0 and received the error.

Baroshem:
I see. Thanks for verification.

One more question if I may.

You have upgraded your app from 1.0.0-rc.2 to 1.0.0 directly or to 1.0.0-rc.3 and received an error?

I am trying to figure out if the issue is in certain RC version or somewhere in the commits for 1.0.0

nicgene:
I upgrading to rc.3 and rc.4 before upgrading to 1.0.0. and could reproduce the issue using releases after rc.2. (Never tried rc.5)

[...]

PR #340, opened 2024-01-12
feat(ssg): disable meta tag for SSG mode

mtdvlpr:
This PR adds a configuration option to disable the meta tag in SSG mode.

## Types of changes

## Description

This is useful when generating HTTP headers using a Nitro Preset, because the meta tag would be redundant and can cause errors in the console, like: `The Content Security Policy directive 'frame-ancestors' is ignored when delivered via a <meta> element.`

## Checklist:

### DIFF
--- a/src/runtime/nitro/plugins/04-cspSsgHashes.ts
+++ b/src/runtime/nitro/plugins/04-cspSsgHashes.ts
@@ -25,7 +25,7 @@ export default defineNitroPlugin((nitroApp) => {
 
     // Parse HTML if SSG is enabled for this route
     if (security.ssg) {
-      const { hashScripts, hashStyles } = security.ssg
+      const { enabled, hashScripts, hashStyles } = security.ssg
 
       // Scan all relevant sections of the NuxtRenderHtmlContext
       type Section = 'body' | 'bodyAppend' | 'bodyPrepend' | 'head'
@@ -102,8 +102,10 @@ export default defineNitroPlugin((nitroApp) => {
     // Generate CSP rules
     const csp = security.headers.contentSecurityPolicy
     const headerValue = generateCspRules(csp, scriptHashes, styleHashes)
-    // Insert CSP in the http meta tag
-    html.head.push(`<meta http-equiv="Content-Security-Policy" content="${headerValue}">`)
+    // Insert CSP in the http meta tag if enabled
+    if (enabled) {
+      html.head.push(`<meta http-equiv="Content-Security-Policy" content="${headerValue}">`)
+    }
     // Update rules in HTTP header
     setResponseHeader(event, 'Content-Security-Policy', headerValue)
 
--- a/docs/content/1.documentation/1.getting-started/2.configuration.md
+++ b/docs/content/1.documentation/1.getting-started/2.configuration.md
@@ -116,6 +116,7 @@ security: {
     exclude: [/node_modules/, /\.git/]
   },
   ssg: {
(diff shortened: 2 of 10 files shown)


src/runtime/nitro/plugins/04-cspSsgHashes.ts
```diff
@@ -102,8 +102,10 @@ export default defineNitroPlugin((nitroApp) => {
     // Generate CSP rules
     const csp = security.headers.contentSecurityPolicy
     const headerValue = generateCspRules(csp, scriptHashes, styleHashes)
-    // Insert CSP in the http meta tag
-    html.head.push(`<meta http-equiv="Content-Security-Policy" content="${headerValue}">`)
+    // Insert CSP in the http meta tag if enabled
+    if (enabled) {
```
Baroshem, line 106: suggestion: I wonder if we shouldn't instead have this check at the `module.ts` level [here](https://github.com/Baroshem/nuxt-security/blob/a6b9c02150ca128240bdfb3af7474988f8b495fc/src/module.ts#L306). 

This way we would be able to not trigger any ssg plugins at all if we are not interested in having them.

Maybe, if you only need to not add meta tag while remain having ssg hashes and scripts, the better option would be to add a `meta` config option instead of `enabled`, WDYT?

Because my understanding of `enabled` boolean flag in `ssg` object would be to disable the work of SSG completely by this module

Baroshem (commented):
Hey, thanks so much for the PR!

It looks great overall, I have only one concern about the naming and the actual usage. Please take a look at the comment I have added in my review :)

src/runtime/nitro/plugins/04-cspSsgHashes.ts
mtdvlpr: Completely agree!

Baroshem:
Hey @mtdvlpr I think that the PR is ready from my point of view. 

The only obstacle right now is the case that was reported in another issue https://github.com/Baroshem/nuxt-security/issues/335.

I will skip merging this PR until this case will be resolved as we want to fix the implemnetation of csp ssg preset first :)

Baroshem (approved):
@mtdvlpr 

I decided that I will approve this PR and merge it even without waiting for the cspPresetPlugin case to be resolved. The ability to disable meta generation can be useful even without csp headers.

@vejja would you like to take a look at this PR? Looks solid from my side :)

Baroshem:
Hey @mtdvlpr 

Sorry but we needed to comment the code for the csp preset plugin. With the latest release of Nuxt, it is throwing errors in this place of the code and we havent figured out yet how to refactor it properly.

For now, you can stay at 1.0.0 version but for the 1.1.0 we are planning to refactor it so that you can still have this functionality but without breaking the code for others in dev and in production.

I think that for now you can skip fixing the tests as once we will refactor the csp plugin preset they will probably work again.

Baroshem:
Hey, no progress so far. I asked for help Nuxt core team member and I am waiting for assistance there.

Regarding hooking to module, have you tried maybe the 1.1.0 branch where we have added support for runtime config? Maybe this could help you

Baroshem:
@mtdvlpr 

Any news?

Baroshem:
Hey @mtdvlpr any news from your side? 

Can I help you somehow to work on this PR? :)

mtdvlpr:
Hi @Baroshem, sorry for the delay! I've been busy. I"ll try to work on this today or tomorrow.

PR #354, opened 2024-01-19
Patch for cheerio performance

### LINKED ISSUE #341: Performance optimization opportunity
### Version

nuxt-security: 1.0.0
nuxt: 3.9.1

### Reproduction Link

https://stackblitz.com/edit/github-3cbcjk

### Steps to reproduce

Go to the StackBlitz, and run `npm run build && npm run preview`. 

And, reload the page. 

By default, the app is not using the `nuxt-security` module. In server log, you will see how long it takes for Nitro to render the app. It averages just ~1ms for me.

To simulate a more real-life page, I am rendering 5000 elements on the page.

Now, enable the `nuxt-security` module and repeat the steps. 

You will see that the Nitro render now averages a lot more now.

### What is Expected?

The security plugins should not impact the Nitro render time that drastically and should scale well with the size of the page.

This seems to be happening in the Nitro plugins that iterate through the HTML context a few number of times using cheerio to get the elements to modify. 

Maybe, we can improve this by reducing the amount of times we loop over the HTML context and making changes as efficiently as possible?

I noticed this issue on a very chunky page. So not sure how much this translates to real world pages. But, nevertheless thought it might be worth investigating if we can improve the performance here.

vejja:
Fixes #341

## Types of changes

## Description

This is a first pass at mitigating cheerio performance issues reported in #341.
Instead of parsing and recombining multiple times the same HTML, we now parse it only once upfront, and recombine it at the end.

It is unlikely to be the best approach as we may prefer to not use Cheerio anymore and instead re-implement via regexes.
I'm proposing this patch as a quick fix in the meantime so that Cloudflare users can mitigate issues.

## Checklist:

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -293,6 +293,15 @@ function registerSecurityNitroPlugins(nuxt: Nuxt, securityOptions: ModuleOptions
       )
     )
 
+    // Pre-process HTML into DOM tree
+    config.plugins.push(
+      normalize(
+        fileURLToPath(
+          new URL('./runtime/nitro/plugins/02a-preprocessHtml', import.meta.url)
+        )
+      )
+    )
+
     // Register nitro plugin to enable Subresource Integrity
     config.plugins.push(
       normalize(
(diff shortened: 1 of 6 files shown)


Baroshem (approved):

Baroshem:
Awesome work @vejja 

Let's merge it for 1.1.0 and test it there and possibly release it this week :)

PR #375, opened 2024-02-13
perf: refactor value manipulations

GalacticHypernova:
## Types of changes

- [x] Enhancement (a non-breaking change which improves functionality like performance)

## Description

Having many manipulations (especially when unnecessary) can be really detrimental to performance. This PR aims to refactor, wherever possible, said iterations.

## Checklist:

Not necessary as it doesn't affect functionality.


GalacticHypernova:
Kept the old code commented for ease of revert in case necessary.

vejja:
@GalacticHypernova 
Let us know if you can achieve measurable performance improvements
Right now CI is failing, will have a look again when you are ready

vejja:
@GalacticHypernova 
CI Output
Capture d’écran 2024-02-14 à 10 45 43
Let me know if you can see the details or if you need a copy of the logs

vejja:
After reviewing the changes, it seems that this PR is mostly re-writing existing code using flow-control statements (`for in`, `for of`, `if else`) instead of functional programming (`entries`, `map`, 'filter`).
It is a matter of personal taste, but I tend to find the functional approach more robust and less buggy.
I don't think the number of loops is reduced with the proposed changes, but happy to change my mind if we can improve performance.

GalacticHypernova:
I will also fix the error of course, it's a bit off CI results appear so late.

GalacticHypernova:
Ah, that makes sense

GalacticHypernova:
> @GalacticHypernova CI Output Capture d’écran 2024-02-14 à 10 45 43 Let me know if you can see the details or if you need a copy of the logs

Would appreciate if I could get the logs for this!

Baroshem:
@GalacticHypernova do you have these benchmarks somewhere to compare performance between the main branch and yours?

I am thinking if it will actually make much difference. I know that in general classic for loops are faster than built in array methods but the difference can be observed mainly when working with datasets that contain milions of rows. 

The optimizations made by @vejja for the cheerio made sense because they were traversing through the generated HTML many times which was decreasing the performance but in the case of your PR it may not even be noticable. Unless there is a benchmark that proves I am wrong :)

GalacticHypernova:
@Baroshem
Thanks, I'm always glad to help! 

I have indeed checked every update I pushed with multiple benchmarks, both on node and on browsers. You're correct it may not be that significant in the short term compared to the cheerio update, but it's an opotimization that will add up at runtime in the long run, benefitting not only the module, but also the users' applications in terms of performance. I can attach some benchmark results if you'd like to see the difference it made.

Baroshem:
Yes please do so!

We will take a look at it with @vejja :)

GalacticHypernova:
Sure thing!
I'll attach them when I get the chance, as I'm not on my PC currently and that's where all my benchmarks are.

GalacticHypernova:
@Baroshem Heya! I really apologize for the long disappearance, got caught up with personal life and haven't had much time to work on this, I have the benchmark results here:
image
image

Here's the benchmark code (ran separately to avoid any sort of engine level caching):

```ts
const bigObject = {}
for(let i = 0; i < 1; i++){
	bigObject[i] = i+1
}
console.time("bigObjectAsArray")
for (let i = 0; i < 1; i++) {
	const bigObjectAsArray = Object.entries(bigObject).map(([key, value]) => [value, key])
}
console.timeEnd("bigObjectAsArray")
console.time("bigObjectAsObject")
for (let i = 0; i < 1; i++) {
	const bigObjectAsArray = []
	for(const key in bigObject){
		bigObjectAsArray.push([bigObject[key], key])
	}
}
console.timeEnd("bigObjectAsObject")
```
As you can see, there is a very clear performance difference even on a single iteration in a single test loops. You could probably imagine how much more noticeable it would really get when it is used on realisting objects (and not ones with a single entry).

GalacticHypernova:
I will also refactor this PR in a new one that I will open soon, because I clearly messed something up right here.

vejja:
Hi @GalacticHypernova 

To be very honest (and I'm not sure you'll like it that much), I am not convinced that we should sacrifice code stability for marginal optimization at this stage of the project. 

There are many core features in the module that are probably going to change significantly (I'm thinking integrating runtime hook events within the core radix router), and I think it would be safer to stabilize these parts of the code before undergoing loop optimization. Although you're right on single line performance timings, we don't loop through that many objects and our objects are not large at all, so overall the impact is not even going to be noticeable.

However, there is one key part of the code where we would really need help in performance/optimization, which is the Cheerio parser. The performance gains are likely to be huge, so if you would accept to have a look at it I think it would be truly fantastic.

Would you have a look at issue #341 ? OP reports that response time went from 14ms to 200-500ms. So if we were able to bring it back down into the 20-50ms range, that would be an amazing 10x performance gain ! In addition we also got reports of CPU/memory crashes in Cloudflare Workers. 

We know exactly what the issue is: Cheerio is very very slow. And we also know what the solution is: use regexes instead of Cheerio. This analysis is based on the fact that OP is comparing Nuxt-Security v0.13 (which used regexes) with Nuxt-Security v1.0 (which uses Cheerio).

Would you by any chance happen to be comfortable with regexes? Let me know your feedback - sorry to be a bit directive here but I think your focus on performance would be highly beneficial to the module.
Cheers

Baroshem:
Hey @GalacticHypernova 

First of all, thanks for the amazing work! You really did a lot of work to benchmark the performance.

But I have to agree on @vejja on this one. I would prefer not to change so many things in the code to achieve such small performance improvement. It is noticable but changing so much can result in unwanted bugs that I would prefer to avoid at all.

But as Vejja mentioned, there are for sure opportunities to significantly improve the prrformance without actually rewriting the code of the module like the example with cheerio. If you would like to cintribute, I am sire this is a great place to start. It would also allow to remove dependency from the module.

This would help the module a lot!

GalacticHypernova:
@vejja @Baroshem 
First things first, thanks for the feedback!
While I would have to disagree about the bug aspect due to the fact tests could be made to avoid regression, I do understand your point. I agree this has less immediate impact on performance (though it does add up at runtime in the long run) and therefore less immediate priority, so I am willing to postpone this until the major issues are taken care of. I'd gladly try to help with cheerio.

vejja:
Closing in favor of #404 by @GalacticHypernova

PR #377, opened 2024-02-15
feat: add escapeHtml option to XssValidator

### LINKED ISSUE #206: Xss validator not allowing having < and > characters in request body

This issue is similar to [Error when sending FormData #147]( https://github.com/Baroshem/nuxt-security/issues/147). The same error is being thrown when there is a request to nuxt api with request body containing `>` and `<` characters. For example: `Organization name>`. It looks like parsing it through the validator causes this problem. Sending `Organization <a>name` is parsed correctly. Disabling xss validator middleware solves this problem but I was wondering if it can accept single `<` or `>` characters without disabling the whole xss validation functionality?

### Version
nuxt-security: v0.14.4
nuxt: v3.5.3

### Steps to reproduce
Send a request to nuxt api with request body containing `<` or `>` or both characters but not a valid HTML tag which are supported.

Morgbn:
resolves #206

## Types of changes

## Description
js-xss has [`escapeHtml`option](https://github.com/leizongmin/js-xss/blob/master/lib/index.js#L18), [by default](https://github.com/leizongmin/js-xss/blob/master/lib/default.js#L158) `<` is replaced by `&lt;` and `>` by `&gt;`
This pull request allows dev to easily disable this feature

## Checklist:

### To test
Before this fail:
```js
fetch('/api/test', { method: 'POST', body: JSON.stringify({ value: "ok ->" }), headers: { 'Content-Type': 'application/json' } })
```
Now, if `xssValidator.escapeHtml` is set to false it works

### DIFF
--- a/src/runtime/server/middleware/xssValidator.ts
+++ b/src/runtime/server/middleware/xssValidator.ts
@@ -1,11 +1,15 @@
-import { FilterXSS } from 'xss'
+import { FilterXSS, IFilterXSSOptions } from 'xss'
 import { defineEventHandler, createError, getQuery, readBody, getRouteRules } from '#imports'
 
 export default defineEventHandler(async (event) => {
   const { security } = getRouteRules(event)
 
   if (security?.xssValidator) {
-    const xssValidator = new FilterXSS(security.xssValidator)
+    const filterOpt: IFilterXSSOptions = { ...security.xssValidator, escapeHtml: undefined }
+    if (security.xssValidator.escapeHtml === false) { // No html escaping (by default "<" is replaced by "&lt;" and ">" by "&gt;")
+      filterOpt.escapeHtml = (value: string) => value
+    }
+    const xssValidator = new FilterXSS(filterOpt)
 
     if (event.node.req.socket.readyState !== 'readOnly') {
       if (['POST', 'GET'].includes(event.node.req.method!)) {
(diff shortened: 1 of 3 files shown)


Baroshem:
Hey @Morgbn 

Great to have you here! Thanks for the pull request. 

I will review it in upcoming days :)

src/runtime/server/middleware/xssValidator.ts
```diff
@@ -1,11 +1,15 @@
-import { FilterXSS } from 'xss'
+import { FilterXSS, IFilterXSSOptions } from 'xss'
 import { defineEventHandler, createError, getQuery, readBody, getRouteRules } from '#imports'
 
 export default defineEventHandler(async (event) => {
   const { security } = getRouteRules(event)
 
   if (security?.xssValidator) {
-    const xssValidator = new FilterXSS(security.xssValidator)
+    const filterOpt: IFilterXSSOptions = { ...security.xssValidator, escapeHtml: undefined }
```
Baroshem, line 8: question: What if user passes `escapeHtml` in `xssValidator`? Wouldn't this line overwrite it automatically to undefined?

Baroshem (commented):
One question as I am not sure

src/runtime/server/middleware/xssValidator.ts
Morgbn: Hey @Baroshem !
Yes, it removes `escapeHtml: boolean` that not compatible with `escapeHtml?: EscapeHandler` ([def](https://github.com/leizongmin/js-xss/blob/8884b21308315915ca5456a8552ddc23366c8325/typings/xss.d.ts#L20))
If `security.xssValidator.escapeHtml` is false, the escapeHtml is set as a function that returns its input directly (which doesn't escapes ">" & "<")
if it's true or undefined, the default function will be used by js-xss (which escapes ">" & "<")

PR #394, opened 2024-03-05
fix(core): do not decode entities in HTML

### LINKED ISSUE #393: Integrate option to disable Cheerio decodeEntities
Hi everybody,

first of all big thanks for this high quality Nuxt module! We have integrated it successfully into a project where we build a news website.

### Problem
Our SEO team has informed us, that since we enabled some Security features (CSP), our server side generated HTML contains altered encoding of certain special characters (e.g. German Umlauts like ä,ö,ü). So we have a difference in SSR HTML and the one which is in the browser.

I debugged a bit and found out that this module uses Cheerio HTML Parsing under the hood. The "problematic" part is https://github.com/Baroshem/nuxt-security/blob/2172055bd282e9b1b19fa033cbecffdd5d66a996/src/runtime/nitro/plugins/02a-preprocessHtml.ts#L22

If we change these options to include `decodeEntities: false` we get the previous behaviour back. 

### Proposition
I understand that it might be a good idea to use the default settings and not change it for all the users. But we would like to include an option somehow to toggle/expose this setting.

### Questions
- Do you know more about the decision to keep the `decodeEntities: true`? Was it a conscious decision?
- Would there be any security implications if we disable the decodeEntities? I do not fully understand how this works together with CSP at all. 

If you need some more info let me know. I could also try to implement the change on my own if you give me some expecation about the config/API changes.

All the best,
Robert ✌️

vejja:
## Types of changes

Fixes #393

Currently, cheerio uses the default setting of `decodeEntities: true` when parsing the HTML document.
This modifies the encoding of special characters (e.g. German umlauts).
This PR changes the option to `decodeEntities: false`, which prevents cheerio from modifying the original source.

## Checklist:

### DIFF
--- a/src/runtime/nitro/plugins/02a-preprocessHtml.ts
+++ b/src/runtime/nitro/plugins/02a-preprocessHtml.ts
@@ -20,6 +20,7 @@ export default defineNitroPlugin((nitroApp) => {
           xml: {
             // Disable `xmlMode` to parse HTML with htmlparser2.
             xmlMode: false,
+            decodeEntities: false
           },
         }, false)
       })


Baroshem:
I will publish a patch version with this fix but I would like to wait as well for this maybe https://github.com/Baroshem/nuxt-security/issues/395#issuecomment-1980370639

Baroshem (approved):

PR #401, opened 2024-03-12
Enhance FormData with XSS validation feature.

### LINKED ISSUE #400: The X-XSS-Protection feature will block FormData
How can I implement XSS protection with FormData because I need to upload files via a form? I've checked the xssValidator.mjs file in nuxt-security/dist/runtime/server/middleware, but it appears to only validate text.

Ray0907:
## Types of changes

- [x ] New feature (a non-breaking change which adds functionality)

## Description
The original version doesn't support FormData, resulting in blocking all FormData requests. This pull request adds a feature to enhance FormData support.
Fix #400 

## Checklist:

### DIFF
--- a/src/runtime/server/middleware/xssValidator.ts
+++ b/src/runtime/server/middleware/xssValidator.ts
@@ -1,29 +1,50 @@
 import { FilterXSS, IFilterXSSOptions } from 'xss'
-import { defineEventHandler, createError, getQuery, readBody, getRouteRules } from '#imports'
+import {
+  defineEventHandler,
+  createError,
+  getQuery,
+  readBody,
+  getRouteRules
+} from '#imports'
 import { HTTPMethod } from '~/src/module'
 
 export default defineEventHandler(async (event) => {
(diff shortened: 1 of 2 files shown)


Baroshem:
Hey @Ray0907 

Thanks for this PR! I will review it in the upcoming days (quite budy days recently)

Ray0907:
@Baroshem 
I'm considering an alternative approach. If I use h3 readMultipartFormData and parse the FormData similar to reading the response body, and only check if the FormData is valid, would that be effective? Because as far as I understand, the XSS validator only checks text. Am I correct? If this approach works, perhaps I'll rewrite the feature

Baroshem:
Hey @Ray0907 

That sounds like an interesting idea. Have you tried how it behaves if a user passes a malicious code in the params?

Like 

`http/localhost:3000?user=<script>`

If it does not break or handles this case properly (it should be validated by XSS as issue and return 400 error) I think we could try this approach :)

PR #404, opened 2024-03-23
perf: avoid cheerio in favor of regex

### LINKED ISSUE #341: Performance optimization opportunity

GalacticHypernova:
## Types of changes

## Description

This PR integrates native solution via regex to load and parse HTML, in order to improve runtime performance.
Closes #341

## Checklist:


src/runtime/nitro/plugins/02a-preprocessHtml.ts
```diff
@@ -10,21 +10,6 @@ export default defineNitroPlugin((nitroApp) => {
...
-          xml: {
-            // Disable `xmlMode` to parse HTML with htmlparser2.
-            xmlMode: false,
-            decodeEntities: false
-          },
-        }, false)
-      })
-    }
-    event.context.cheerios = cheerios
+    event.context.cheerios = html
```
GalacticHypernova, line 13: @Baroshem I'm not entirely sure, should this be trimmed? Is there anything that needs to be done with each cheerio? Or is it meant to be as is?

Baroshem: I think @vejja would be a best person to answer here as he was implementing the cheerio :)

Baroshem: But I think that the best case would be to get rid of all library in general and use regexes instead for better performance :)

GalacticHypernova: > But I think that the best case would be to get rid of all library in general and use regexes instead for better performance :)

I apologize, it appears I misspoke. I meant to ask, what should each html section receive? I referred to a section as cheerio because the context is cheerios, I apologize for the misunderstanding.

GalacticHypernova:
@vejja thanks for the explanation! This really did help a lot, and will be very useful for the PR. Does that means no further modification needs to be made for 02a? No need to trim or anything like that? I was just wondering if it actually needs to be modified at all.

vejja:
> @vejja thanks for the explanation! This really does help a lot, and will be very useful for the PR. Does that mean no further modification needs to be made for sections in 02a (before the 3 routines)? No need to trim or anything like that? I was just wondering if it actually needs to be modified at all. P.S. I already have an idea in mind to handle caching these so that there won't be a need to iterate over everything again.

I think if you can get rid of 02a and 99b, it would be fantastic. Honestly they are a legacy of trying to quick-fix the cheerio issues

GalacticHypernova:
> I think if you can get rid of 02a and 99b, it would be fantastic. Honestly they are a legacy of trying to quick-fix the cheerio issues
👍 
I mean, I do have a caching strategy in mind which might require 02a, I'm not sure yet, but 99b I do believe will be unnecessary.

GalacticHypernova:
If I am correct (I haven't worked much with nitro's HTML hooks myself), the bodyAppend/prepend/etc return strings, right? They just return the stringified HTML? Because I see a `$('script')` inside a forEach so I want to make sure

vejja:
I can’t remember but I think Typescript gives you the type

GalacticHypernova:
Thanks! I remembered that vaguely but I also didn't get the chance to check it myself.

GalacticHypernova:
@vejja  in 03, can external scripts have `<script src=...></script>` syntax, or is it guaranteed to be `<script src=.../>`

vejja:
> @vejja in 03, can external scripts have `<script src=...></script>` syntax, or is it guaranteed to be `<script src=.../>`

I think in theory it should always be the first syntax as <script> is not supposed to be self-closing. So my mistake for suggesting the other way round.
However there used to be many issues around that, so I guess it’s still possible in theory to have <script …/>

vejja:
Also if you're wondering what regex we used to have before Cheerio, you can check out:
https://github.com/Baroshem/nuxt-security/blob/v1.0.0-rc.1/src/runtime/nitro/plugins/02-cspSsg.ts
https://github.com/Baroshem/nuxt-security/blob/v1.0.0-rc.1/src/runtime/nitro/plugins/99-cspNonce.ts

GalacticHypernova:
> > @vejja in 03, can external scripts have `<script src=...></script>` syntax, or is it guaranteed to be `<script src=.../>`
> 
> I think in theory it should always be the first syntax as <script> is not supposed to be self-closing. So my mistake for suggesting the other way round. However there used to be many issues around that, so I guess it’s still possible in theory to have <script …/>
> 
> **Update:** Upon reasearching the topic, I think we should **always** assume it is going to be `<script ...></script>`.
> 
> Let's keep our lives simple and not support `<script .../>`. It is not HTML spec-compliant, and Nuxt does not use this syntax. Trying to have a regex that catches a useless scenario is likely to introduce more bugs than solve problems.
> 
> Thanks for bringing this up @GalacticHypernova

Thanks!
In theory I could still support that syntax relatively easily, it depends on whether nitro handles the html conversion to the spec or not, otherwise user-supplied scripts may not have the integrity if they do use the self-closing syntax.

vejja:
> Thanks! In theory I could still support that syntax relatively easily, it depends on whether nitro handles the html conversion to the spec compliant version or not, otherwise user-supplied scripts may not have the integrity if they do use the self-closing syntax.

Up to you. You're absolutely right that otherwise the won't have integrity if they use the self-closing syntax

src/runtime/nitro/plugins/03-subresourceIntegrity.ts
```diff
@@ -2,6 +2,8 @@ import { useStorage, defineNitroPlugin, getRouteRules } from '#imports'
 import { isPrerendering } from '../utils'
 import { type CheerioAPI } from 'cheerio'
 
+const SCRIPT_RE = /<script((?=[^>]+src="[\w:.-\\/]+")(?![^>]+integrity="[\w-]+")[^>]+)\/>/g
```
GalacticHypernova, line 5: Progress report on the script regex:
```ts
const SCRIPT_RE = /<script((?=[^>]+src="[\w:.-\\/]+")(?![^>]+integrity="[\w-]+")[^>]+)(?:\/>|><\/script>)/g

const a =[
  `<script src="a.k"/><script src="b"></script>`,
  `<div>`,
  `<script>c</script>`,
  `<script src="a" integrity="a"></script> <script src="b"/>`
]
console.log(a.map(r=>r.replace(SCRIPT_RE,(match, mb)=>{return `<script integrity="a"${mb}></script>`})))
```
image


It works for both self-closing and non-self-closing script tags and automatically converts them to non-self-closing.

GalacticHypernova:
Also what about links? Are they always self closing? Or should I treat them the same as script tags?

src/runtime/nitro/plugins/03-subresourceIntegrity.ts
vejja: Very nice !
Can you test also for
- integrity attribute before src attribute
- random attribute (e.g. `data-xxx`) before and after src attribute
- no src attribute with self-closing syntax
As well as edge cases such as :
- `<script>console.log('I want to mess up with javascript by printing </script>')</script>`
- `<p> I like to mess up with pre tags such as <pre> a code example: <script></script> </pre> </p>`
- `<script></script>

vejja:
> Also what about links? Are they always self closing? Or should I treat them the same as script tags?

I think `<link>` is always empty. It is never self closing, and it has no closing `</link>` tag.
So it should always be like `<link rel="stylesheet" href="example.css">`

src/runtime/nitro/plugins/03-subresourceIntegrity.ts
GalacticHypernova: Will do and post pictures soon!

GalacticHypernova: image


Tested with this sample:
```ts
const SCRIPT_RE = /<script((?=[^>]+src="([\w:.-\\/]+)")(?![^>]+integrity="[\w-]+")[^>]+)(?:\/>|><\/script>)/g
const sriHashes = {
  "a.k":"a",
  "b":"c",
  "script.js":"newIntegrity"
}
const a = [
  `<script src="a.k"/><script src="b"></script>`,
  `<div>`,
  `<script>c</script>`,
  `<script src="a" integrity="a"></script> <script src="b"/>`,
  `<script integrity="integrity" src="script.js"></script>`, //integrity before src
  `<script data-xxx src="script.js"></script>`, //random attr before
  `<script src="script.js" data-yyy></script>`, //random attr after
  `<script data-xxx src="script.js" data-yyy></script>`, //random attr before and after
  `<script src="script1.js"></script><script src="script2.js"></script>`, //multiple scripts in the same line
  `<script>console.log('I like to mess up with javascript by printing </script>')</script>`, //first edge case - print
  `<p> I like to mess up with pre tags such as <pre> a code example: <script></script> </pre> </p>`, // second edge case - <pre>
  `<p> I like to mess up with pre tags such as <pre> a code example: <script> that is non-closing by mistake </pre> </p>`, // second edge case - <pre> (evil version)
  `<script><!-- I like to mess up with HTML comments such as </script> --></script>`, // third edge case - comment inside script
  `<script><!-- I like to mess up with HTML comments such as <script> --></script>` // third edge case - comment inside script (variant)
];
console.log(a.map(r=>r.replace(SCRIPT_RE,(match, rest, src: Partial<keyof typeof sriHashes>)=>{
  const hash = sriHashes[src]
  if (hash) {
    return `<script integrity="${hash}"${rest}></script>`
  }
  return match
})))
```
        
Please let me know if anything got incorrect results!

vejja: Amazing, thanks a lot !!

GalacticHypernova: Glad to help! Of course I will also add those tests to e2e tests in the repo

GalacticHypernova: The real issue would be the inline scripts, I would need to think about how to make the regex robust enough.

GalacticHypernova: As per the cache, what do you think about using a `cache` key in the event.context? A use case for that would be the fact that for external scripts in 04, for example, we won't need to re-test it with a regex. We could save a WeakMap when we go through 03, where each script string would have the integrity as the value. That way we could simply get it from the map in stead of making even more regex operations. Is there anything that could hinder it as far as you know?

GalacticHypernova: Also, one more thing. Would it be too much of an issue if "valid" external scripts would get integrity? I just tested with `console.log(<script src="a.k"></script>)` and it returned the the script with the integrity, but I don't really think that would cause any issues, would it?

vejja: Sounds good to me

GalacticHypernova: Yea, unfortunately you can't win them all with regex.. But at least the edge cases and potential issues are covered, so that's something

vejja: > Also, one more thing. Would it be too much of an issue if "valid" external scripts would get integrity? I just tested with `console.log(<script src="a.k"></script>)` and it returned the the script with the integrity, but I don't really think that would cause any trouble, would it?

I am not too sure to be honest
- On the one hand, in theory this is a problem that could lead to many issues. We are modifying user code by doing XSS ourselves here, which looks really bad for a security module
- On the other hand, I don't even know how browsers are supposed to interpret this... 

This might be the reason why the author of https://github.com/Baroshem/nuxt-security/blob/v1.0.0-rc.1/src/runtime/nitro/plugins/99-cspNonce.ts was doing some lookbehind search around quotes

GalacticHypernova:
We might want to look into a hybrid approach, at least until a suitable solution is found for the edge cases that are harder to catch with regex. That is, for external scripts/links/etc, regex would be used, but for inline stuff cheerio. That way it would still increase performance (as it would be used less), but wouldn't compromise integrity entirely (again, as a temporay solution).
Would like to hear your thoughts on this!

src/runtime/nitro/plugins/03-subresourceIntegrity.ts
vejja: > Yea, unfortunately you can't win them all with regex.. But at least the edge cases and potential issues are covered, so that's something

Sorry, I was commenting on your idea of a WeakMap cache
Our comments on integrity injection crossed at the same time

vejja:
> We might want to look into a hybrid approach, at least until a suitable solution is found for the edge cases that are harder to catch with regex. That is, for external scripts/links/etc, regex would be used, but for inline stuff cheerio. That way it would still increase performance (as it would be used less), but wouldn't compromise integrity entirely (again, as a temporay solution). Would love to hear your thoughts on this

Isn't an inline script the same as an external script except that it doesn't have an src attribute ?
Let me dig into this, I need to read the HTML spec 🤮🤮🤮

src/runtime/nitro/plugins/03-subresourceIntegrity.ts
GalacticHypernova: I see, all good!

GalacticHypernova: > This might be the reason why the author of https://github.com/Baroshem/nuxt-security/blob/v1.0.0-rc.1/src/runtime/nitro/plugins/99-cspNonce.ts was doing some lookbehind search around quotes

Yea, but at the same time it could be something like:
<div> hi'<script>console.log('hi')</script></div>

GalacticHypernova: It's not that straight-forward unfortunately

vejja:
> > Isn't an inline script the same as an external script except that it doesn't have an src attribute ? Let me dig into this, I need to read the HTML spec 🤮🤮🤮
> 
> Well, not exactly. Assuming the worst case scenario, that everyone using this module is trying to cause harm, there are a lot more ways for inline scripts to cause trouble.

I'm not sure what the issue could be ?

GalacticHypernova:
> I'm not sure what the issue could be ?

<script>console.log('me like </script>')</script><script>hi</script>

GalacticHypernova:
I mean, this would be x10 easier if the html's sections are actually one element each. I'll try it out real quick

GalacticHypernova:
@vejja I have discovered some things:
1. It appears as though each section returns a single string in an array (it appears as though they are concatenated with `\n` and `+`).
image

2. Script elements cannot be nested, which is great, as that takes away most of the complexities with checking script within script.
3. Unfortunately the script contents don't have their own dedicated spot.
4. It appears as though my comment about the </script> inside console.log and the likes is not even possible, which is also great as that takes away some more complexities (tested in the <template>). This should get tested a bit more, but this means that we can safely assume </script> means the actual end of the script.
image

vejja:
> @vejja I have discovered some things:
> 
> 1. It appears as though each section returns a single string in an array (it appears as though they are concatenated with `\n` and `+`).

I think at least in head there can be several items in the array. 
Probably also depends on Nuxt settings such as  nuxt.config.ts and useHead parameters ?

> 2. Script elements cannot be nested, which is great, as that takes away most of the complexities with checking script within script.
> 3. Unfortunately the script contents don't have their own dedicated spot.
> 4. It appears as though my comment about the </script> inside console.log and the likes is not even possible, which is also great as that takes away some more complexities (tested in the template), as it errors stating "invalid end tag", and also translates to dev app startup failure. This should get tested a bit more (like production env), but this probably means that we can safely assume </script> means the actual end of the script.

Very interesting, seems that your regex is good then ?

GalacticHypernova:
> Very interesting, seems that your regex is good then ?

I mean, I'm not entirely sure yet, becasue as I said, this still needs to get thoroughly tested to not have its own edge cases, but from what I had gathered so far, through layouts/app.vue/pages, both in templates and through useHead for programmatic script insertion, it errors on both build and dev, but I only tested it for like 20 minutes so I would try to test it a bit further before making the final verdict, but I guess for the most part this means the regex is indeed good

GalacticHypernova:
@Baroshem you're probably way more experienced in nuxt than me, could I get a confirmation that doing anything with "</script>" inside a script is disallowed and will fail to compile?

vejja:
> @Baroshem you're probably way more experienced in nuxt than me, could I get a confirmation that doing anything with "</script>" (like in a console.log) inside a script is disallowed and will fail to compile, meaning there is no special handling to add to such case?

Hi @GalacticHypernova 
I just verified, and it is not possible to include `</script>` in a javascript string. Trying to do so triggers a `unterminated string literal` Javascript error.

Baroshem:
Hey guys, sorry but I was wuite busy for the last few days.

If you need my opinion on something please let me know and I will try to answer the best I can :)

GalacticHypernova:
> I just verified, and it is not possible to include `</script>` in a javascript string. Trying to do so triggers a `unterminated string literal` Javascript error.

Thank you! This does very much simplify it.

src/runtime/nitro/plugins/04-cspSsgHashes.ts
```diff
@@ -43,13 +56,9 @@ export default defineNitroPlugin((nitroApp) => {
               if (!src && scriptText) {
                 // Hash inline scripts with content
                 scriptHashes.add(`'${generateHash(scriptText, hashAlgorithm)}'`)
```
GalacticHypernova, line 58: @vejja I have a pattern in mind for inline scripts, but I don't know how cheerio works in this case:
Do we need to make sure the contents of the script is non-empty-ish? That is, no whitespace only? 
Because the pattern would change accordingly;
If it's only that it's not empty, this can be used:
```ts
/<script(?![^>]*?src="[\w:.-\\/]+")>(.*?)<\/script>/g
```
If it's that it's not whitespace only, this can be used:
```ts
/<script(?![^>]*?src="[\w:.-\\/]+")>(?!\s*?<\/script>)(.+?)<\/script>/g 
```
The difference is the negative lookahead in the second pattern to make sure it's not only whitespace until the closing tag, which entirely depends on cheerio's functionality in such cases.

GalacticHypernova:
@vejja @Baroshem this is still not done, however I wanted to ask what your opinions on the current implementation is, and if theres' anything to change.

src/runtime/nitro/plugins/04-cspSsgHashes.ts
vejja: Hi @GalacticHypernova 
I need to do a full review, but in the meantime I just want to understand your question above.
1- inline scripts don't have the `src` attribute
2- external scripts have the `src` attribute. They can have an `innerHtml` content but it will be ignored
Cheerio parses the `src` attribute and the `innerHtml` node separately. These are two different objects, located in two different places.

vejja:
> @vejja @Baroshem this is not yet finished, however I wanted to ask what your opinions on the current implementation is, and if there's anything I should change (the errors in the tests don't seem quite right considering I'm directly modifying them as strings and return the modifies ones so I don't quite know where the origin of the error is).

My opinion is that it looks good, thanks !!

The first test that fails is `it('injects `integrity` on Nuxt root scripts')`
In that test, the section `head` has only one array item, and it is a string which is roughly in this format :

```html
<meta charset="utf-8">
<meta name="viewport" content="width=device-width, initial-scale=1">
<link rel="modulepreload" as="script" crossorigin href="/_nuxt/entry.DYpycOZ-.js">
<link rel="modulepreload" as="script" crossorigin href="/_nuxt/index.-yXp0hiT.js">
<link rel="modulepreload" as="script" crossorigin href="/_nuxt/nuxt-link.sHJ0UhRl.js">
<link rel="prefetch" as="style" href="/_nuxt/error-404.c7Cl-q-2.css">
<link rel="prefetch" as="script" crossorigin href="/_nuxt/error-404.DFp4BJVl.js">
<link rel="prefetch" as="script" crossorigin href="/_nuxt/vue.f36acd1f.DbAQkOB5.js">
<link rel="prefetch" as="style" href="/_nuxt/error-500.DacwWOBj.css">
<link rel="prefetch" as="script" crossorigin href="/_nuxt/error-500.DKZOBwlI.js">
<script type="module" src="/_nuxt/entry.DYpycOZ-.js" crossorigin></script>
```

The `SCRIPT_RE` and `LINK_RE` regexes in `04-subresourceIntegrity` don't find the correct strings to inject `integrity` attributes, which are in lines 3, 4, 5 and 11 (respectively: link to entry.js, link to index.js, link to nuxt-link.js, and script module entry.js).

GalacticHypernova:
> Update: this is due to the presence of minus (`-`) hyphen in the `src` and `href` strings, which is excluded by the `w` regex check.

Ah I see that, yea, sorry! I missed that one, the isssue was it defaulted to regex as I forgot to escape the `-`

src/runtime/nitro/plugins/04-cspSsgHashes.ts
GalacticHypernova: Alright, let's keep it as is then

GalacticHypernova:
Update: I fixed some of the tests now, there was a slight misconfiguration in the regex. I'm still looking into the ones left

vejja:
> > I think you can delete the w check, as anything between the two " quotations mark would fit.
> 
> Well, I could, but I thought it would be better to provide a more focused pattern to prevent as many possible edge cases, in stead of resorting to something generic like `.`

I just checked and file names can include any UTF-8 character (including accents such as `éùö`, symbols such as `€§` and Chinese characters such as `上海+中國`). The list would be too long to assemble properly.
So let's go for the generic `.`
With probably the benefit of being faster ?

GalacticHypernova:
Well, in that case I will rework it. We'll just need to add thorough tests to make sure there are no regressions.

Baroshem:
Hey folks, any progress on this? I would love to see this being added to the upcoming 2.0.0 release :)

GalacticHypernova:
Hey! Sorry for the slow progress, been very busy with real life matters. I'm currently trying to debug the errors, I believe those are the only blockers. If you have any idea why they may be popping up I could push a fix, maybe even today.

GalacticHypernova:
@vejja do you have any idea why the tests are failing? Nothing in this PR seems to be the cause of this. Is the issue in the unit tests?

vejja (approved):
LGTM now
Only substantial modification was to make 'as' optional in LINK_RE when scanning ssgHashes

GalacticHypernova:
@vejja glad I could help! I just noticed I accidentally made as required when making the pattern, sorry about that and thanks for catching it!

GalacticHypernova:
I have 2 main candidates for the optional as pattern:
We can use `/<link(?=[^>]+\brel="(stylesheet|preload|modulepreload)")(?=[^>]+\bintegrity="([\w\-+/=]+)")(?=[^>]+\bas="(\w+)"|)[^>]+>/`
But we can also use the pattern @vejja suggested, which is `/<link(?=[^>]+\brel="(stylesheet|preload|modulepreload)")(?=[^>]+\bintegrity="([\w\-+/=]+)")(?=(?:[^>]+\bas="(\w+)")?)[^>]+>/`

Any preferences? The only difference is that in the first pattern the positive lookahead adds an empty alternative with `|`, whereas the second one makes the inner pattern optional, so it would essentially do the same and check for either the as attribute or an empty alternative.

vejja:
Can you give an example where the 2 patterns would give a different result ?

GalacticHypernova:
> Can you give an example where the 2 patterns would give a different result ?

I can try to find some edge case but from what I have tested thus far its the same, only expressed differently. The first pattern essentially tells the regex engine to make sure there is either the as attribute or an empty alternative (which would always be true since it's fallback), and the second pattern (the one you suggested) tells the engine to look for an a potential match with the as attribute, meaning it will also accept nothing as it's only a potential match.

PR #408, opened 2024-04-01
fix: set nonce in response headers instead of using configuration

### LINKED ISSUE #407: Cannot use nonce in nuxt-security:headers hook
### Version

nuxt-security: 1.2.2
nuxt: 3.11.1

### Reproduction Link

https://stackblitz.com/edit/nuxt-security-missing-nonce?file=server%2Fplugins%2Fsecurity.ts

### Steps to reproduce

1. Configure the `nuxt-security:headers` hook to change CSP with Runtime Config [as per docs](https://nuxt-security.vercel.app/documentation/getting-started/usage#runtime-configuration)
2. Include a `'nonce-{{nonce}}'` value in one of the CSP values

```typescript
  nitroApp.hooks.hook('nuxt-security:ready', async () => {
    await nitroApp.hooks.callHook('nuxt-security:headers', {
      route: '/**',
      headers: {
        contentSecurityPolicy: {
          'script-src': [
            "'self'",
            'https:',
            "'unsafe-inline'",
            "'strict-dynamic'",
            "'nonce-{{nonce}}'",
          ],
        },
      },
    });
  });
});
```

### What is Expected?

The CSP header should be transformed to include the actual nonce value.

The CSP value contains the exact string `'nonce-{{nonce}}'`

image

By adding some logging into the `nuxt-security` nitro plugins files, you can see that the custom CSP from the hook is being registered. However, when the plugin `99-cspSsrNonce.ts` runs, the CSP value it resolves and replaces the nonces in, is the default/nuxt.config.ts CSP.

https://github.com/Baroshem/nuxt-security/blob/13a96a6e36989a277cb046b379bb65a251e04afc/src/runtime/nitro/plugins/99-cspSsrNonce.ts#L43

This is meaning we cannot do a combination of using runtime config in the CSP and nonces.

huang-julien:
## Types of changes

## Description

Fix #407 

Hey :wave: 
This PR change the way the SSR CSP plugin works. Before that, it was relying on the configuration. 

This Pr moves the replacement of the nonce placeholder to `beforeResponse`, so it can parse and rewrite the content security policy header after it was set (by runtime hooks and the security configuration/build-time config)

## Checklist:

### DIFF
--- a/src/runtime/nitro/plugins/99-cspSsrNonce.ts
+++ b/src/runtime/nitro/plugins/99-cspSsrNonce.ts
@@ -1,7 +1,7 @@
 import { defineNitroPlugin, getRouteRules, setResponseHeader } from '#imports'
 import { type CheerioAPI } from 'cheerio'
 import { isPrerendering } from '../utils'
-import { H3Event } from "h3"
+import type { H3Event } from "h3"
 
 export default defineNitroPlugin((nitroApp) => {
   nitroApp.hooks.hook('render:html', (html, { event }) => {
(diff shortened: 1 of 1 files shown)


Baroshem:
Thank you so much @huang-julien for creating this PR! 💚

I will merge it soon and release it with upcoming 1.3.0 version

PR #410, opened 2024-04-03
docs: use new `nuxi module add` command in installation

danielroe:
## Types of changes

## Description

This updates the documentation to use the [`nuxi module add` command](https://github.com/nuxt/cli/pull/197) which should simplify docs a bit and also improve user experience as there's no need to add to `nuxt.config` manually.

It's documented [here](https://nuxt.com/docs/api/commands/module#nuxi-module-add).

I may have missed a few spots in the documentation as I'm doing this across the modules ecosystem assisted by the power of regular expressions ✨, so I'd appreciate a review 🙏

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,7 @@ Automatically configure your app to follow OWASP security patterns and principle
 Install the module:
 
 ```sh
-npm i -D nuxt-security
-yarn add -D nuxt-security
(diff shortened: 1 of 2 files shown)


Baroshem:
Thank you so much @danielroe ! 🚀

PR #435, opened 2024-04-25
feat(core): use virtual file system for SRI

### LINKED ISSUE #395: Subresource Integrity error
### Version

nuxt-security: 1.2.1
nuxt: 3.10.3

### Reproduction Link

n/a

### Steps to reproduce

n/a

### What is Expected?

No SRI error after deploying to Azure Static Web Apps with `azure` preset with default Nuxt Security's SRI config.

image

vejja:
## Types of changes

## Description

This PR brings a technical rewrite of the way we save SRI hashes at build time.

Previously, the way we saved SRI hashes was hacky. We were writing the hashes on the drive under the `.nuxt` build-time directory, which was necessary for SSG rendering of integrity hashes. But because this directory is not available in the server context in SSR mode, we also copied them to the server assets output folder. Then we relied on `useStorage` to retrieve the hashes, and depending whether the context was SSR or SSG, we had to look either into the `build` folder or `assets` folder. Because the output was in JSON format, we encountered issues when Nitropack changed the format of raw assets in JSON format (most notably #395).

With this rewrite, we now use the native Nuxt Virtual File System to inject SRI hashes at build time. In practice, this means that the SRI hashes are now directly bundled into the Nitro server runtime. We do not need `useStorage` to read them from the disk anymore.

This closes #395 definitively as we remove the dependency from `useStorage` raw format.

## Checklist:

### DIFF
--- a/playground/nuxt.config.ts
+++ b/playground/nuxt.config.ts
@@ -2,6 +2,7 @@ import { defineNuxtConfig } from 'nuxt/config'
 
 export default defineNuxtConfig({
   modules: ['../src/module'],
+  devtools: { enabled: true },
 
   // Per route configuration
   routeRules: {
@@ -46,6 +47,7 @@ export default defineNuxtConfig({
   // Global configuration
   security: {
     headers: {
+      crossOriginEmbedderPolicy: false,
       xXSSProtection: '0'
     },
     rateLimiter: {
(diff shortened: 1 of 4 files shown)


Baroshem (approved):
Looks really good, thanks! Do you think we could release it as a patch version or should we wait until the next minor?

vejja:
It can be a patch, this is really small

Baroshem:
Published as 1.4.3 :)

Issue #348, opened 2024-01-17
"This module cannot be imported in server runtime" error on fresh Nuxt install


alkstal:
### Version

nuxt-security: 1.0.0
nuxt: 3.9.3

### Reproduction Link

https://github.com/johanalkstal/nuxt-security-repro

### Steps to reproduce

Run `pnpm install`
Run `pnpm run dev`

### What is Expected?

No errors

### What is actually happening?

```
Nuxt 3.9.3 with Nitro 2.8.1                                                                                 3:41:14 PM
                                                                                                            3:41:14 PM
  ➜ Local:    http://localhost:3000/
  ➜ Network:  use --host to expose

  ➜ DevTools: press Shift + Alt + D in the browser (v1.0.8)                                                 3:41:15 PM

ℹ Vite server warmed up in 870ms                                                                           3:41:17 PM
ℹ Vite client warmed up in 1043ms                                                                          3:41:17 PM

[3:41:17 PM]  ERROR  This module cannot be importedin server runtime. [importing @nuxt/kit from node_modules/.pnpm/nuxt-security@1.0.0/node_modules/nuxt-security/dist/runtime/nitro/plugins/05-cspSsgPresets.mjs]
```

vejja:
@Baroshem 
This is the same issue with importing `@nuxt/kit` in the nitro plugin
@pi0 has now implemented guards against this in Nuxt 3.9.3
Maybe we need to temporarily disable `05-cpsSsgPresets` while we look for an alternative ?

Baroshem:
Hey @vejja 

Yes, it is related to that.

Could you create a fix for that? I will publish a patch version later today that we will fix with 1.1.0

nicokempe:
Thank you for your quick action 🙏🏽

ethikz:
@Baroshem It is gone for me now. Thank you!

Issue #438, opened 2024-04-26
RollupError: Could not resolve `../../../utils`


schaubes:
After todays update to v1.4.0 using pnpm I get following error:
```
ERROR  RollupError: Could not resolve "../../../utils" from "node_modules/.pnpm/nuxt-security@1.4.0_rollup@4.16.4/node_modules/nuxt-security/dist/runtime/nitro/plugins/00-routeRules.mjs"
```

### Version

nuxt-security: v1.4.0
nuxt: v3.11.2

### Reproduction Link

Can post if necessary

### Steps to reproduce

Initializing a new nuxt app with pnpm and adding nuxt-security

```bash
npx nuxi@latest init test-app
cd test-app/

# add nuxt-security
npx nuxi module add security

# run dev server with pnpm
pnpm run dev
```

### What is Expected?

`pnpm run dev` should start normally

### What is actually happening?

```
ERROR  RollupError: Could not resolve "../../../utils" from "node_modules/.pnpm/nuxt-security@1.4.0_rollup@4.16.4/node_modules/nuxt-security/dist/runtime/nitro/plugins/00-routeRules.mjs"
```

MarijnFK:
Jup, same here

Baroshem:
Hey, thanks for reporting this issue. We are on it!

Baroshem:
This is really strange. At first I thought that the issue is related to wrong dist generation (I copied the dist to the node modules of the project) and project started working.

But it can be something else, still investigating.

vejja:
> It should work now with 1.4.2. @MarijnFK @schaubes could you verify?
> 
> @vejja I needed to add a copy of `defuReplaceArray` to the utils of the runtime. The import works locally but fails when released in a published dist package.

Ah yes makes sense now

garyellow:
> It should work now with 1.4.2. @MarijnFK @schaubes could you verify?
> 
> @vejja I needed to add a copy of `defuReplaceArray` to the utils of the runtime. The import works locally but fails when released in a published dist package.

It works with v1.4.2
thx~~

MarijnFK:
Quickly fixed, thanks!

PR #441, opened 2024-04-28
feat(core): Security Headers for Pre-rendered Routes

### LINKED ISSUE #386: [Feature Request]: expose generated security headers at build time
I use SSG and I want to be able to extract the generated security headers (especially the csp with defined integrity hashes), so that I can write those headers to an nginx configuration file before deploying the static build output to my nginx server.

I'm not sure how this feature can be implemented best, but here are some vague ideas I have:

- Fix [05-cspSsgPresets.ts](https://github.com/Baroshem/nuxt-security/blob/main/src/runtime/nitro/plugins/05-cspSsgPresets.ts) somehow
- Write the headers to a file so a user can read the file and use it's content
- Provide a hook that exposes the headers once they are generated for a user to hook into (similar to the nitro compiled hook)

I tried to use the runtime hooks feature introduced in v1.1.0, but this didn't work.

Another solution for me would be to calculate the integrity hashes myself in some script and then use those values to create my header configuration, but it makes more sense to use this module, because it also forces the header values in dev mode, so they can be tested and it gives some common sense defaults that I wouldn't have to configure manually.

### Additional context
Related to #336.

### LINKED ISSUE #434: Support for Security Headers for resources other than HTML
Hello,

I've noticed that all web security headers only appear in the HTML responses, but not in the Nuxt resources such as CSS, JS, etc. Public resources are also not considered by Nuxt-Security.

Security scanners are raising alarms because the necessary headers, such as X-Content-Type-Options: nosniff, are not set on the resources.

Currently, I've solved this issue with an additional proxy that sets these headers on all resources. However, I suspect this might be a bug in Nuxt-Security, but I'm not sure as I'm not very familiar with this area.

Is it intended that the security headers are only set in the HTML responses? Or is this a bug?

Thank you in advance for your help.

Best regards

vejja:
## Types of changes

## Description

With this PR, security headers
- are now available in Hybrid Applications (see #409)
- are also available in Client-Side Rendered Applications via Nitro presets or hook (see #386)
- are available on non-HTML resources, where relevant (see #434)

Previously, when a route was pre-rendered via `routeRules`, the Nuxt application could not deliver the Security Headers. This is because the HTML page is served directly as a pre-built static asset, and we could not intercept the headers in that case to modify them.

This PR proposes a modification of the core engine of Nuxt Security, whereby we now record all pre-rendered pages at build time, and then intercept all requests to verify whether they correspond to a pre-rendered page.

As a consequence of this core upgrade, we now deliver five additional features:
1. For pre-rendered pages in a Nuxt Hybrid application, we are now able to provide **CSP by way of headers**, in addition to the `meta` tag. This addresses the issue where some security scanners are complaining about CSP compliance because they do not check the `meta` tag.
2. More generally, we are now able to provide _all_ security headers for pre-rendered pages. Previously, only CSP could be provided through the `meta` tag, but now **all other policies** (e.g. COEP, STS, etc...) can also be delivered via headers for pre-rendered pages in Hybrid Mode.
3. We take this opportunity to also re-activate the `cspSsgPresets` feature, which had to be deactivated after the upgrade to Nuxt 3.9.3 caused issue #348. With this feature, we now generate a `headers.json` file that is used by Nitro in static presets. For instance, a **static website** deployed on Vercel or Netlify now has security headers natively on all pages, even without a server runtime !
4. We introduce a new `nuxt-security:prerenderedHeaders` build-time hook, which allows to further extend the previous feature **when there is no Nitro preset available**. This hook can be used to manually generate a headers configuration file for any custom static deployment.
5. In addition, we now add some common security headers on **non-HTML assets**, such as images, `.js` or `.css` files, api requests, etc. The following headers are now delivered on all resources: `referrerPolicy`, `strictTransportSecurity`, `xContentTypeOptions`, `xDownloadOptions`, `xFrameOptions`, `xPermittedCrossDomainPolicies`, and `xXSSProtection`.

Please note that : 
- The features 1. and 2. are available under a new `ssg: nitroHeaders` option (default `true`). This option is modifiable both globally and at route-level.
- Feature 3. is available under a new `ssg: exportToPresets` option (default `true`). This option is only available globally.
- Features 4. and 5. are always available (no option switch).

Closes #386 
Closes #409 
Closes #434

## Checklist:


Baroshem:
Hi @vejja 

Thanks for this amazing PR!

I needed to first release your other PR with SRI fix and it caused the conflicts in this PR. Could you please rebase it? :)

Also, regarding the release of this feature, I would aim for 1.5.0 next week probably if that is ok with you? :)

I will also review it in the upcoming days to give you some feedback.

vejja:
I'm also closing #434 with this PR
Will update description and notes

vejja:
@Baroshem 
As a side technical note, I am also fixing the issue that popped up with #438, where we had to duplicate some `utils` in different folders because runtime code could not access files in the `src` folder.
I am adding a `build.config.ts` file that tells `unjs/unbuild` to make the `utils` code available at the root. 
This allows to put all common code in `utils`, no duplication required anymore !

build.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineBuildConfig } from "unbuild"
+export default defineBuildConfig({
+  declaration: true,
+  entries: [
+    'src/utils/hash.ts',
+    'src/utils/headers.ts',
+    'src/utils/merge.ts'
+
```
Baroshem, line 8: nitpick: unnecessary break line

playground/nuxt.config.ts
```diff
@@ -55,6 +56,24 @@ export default defineNuxtConfig({
       interval: 30000,
       headers: true
     },
-    removeLoggers: false
+    removeLoggers: false,
+  },
+
+  hooks: {
+    'nuxt-security:prerenderedHeaders': async(prerenderedHeaders) => {
+      console.log('in prerenderedHeaders hook', prerenderedHeaders)
```
Baroshem, line 64: question: do we want this console.log to be here?

Baroshem (approved):
Two small nitpicks, overall really great work!

Baroshem:
Amazing work @vejja !

I will merge it to the 2.0.0-rc.1 branch where we can do some additional testing and prepare a big release soon :)

PR #452, opened 2024-05-13
fix(core): spread storage options

### LINKED ISSUE #433: Rate limiter driver options not being properly passed
### Version

nuxt-security: V1.3.2
nuxt: v3.11.2

### Steps to reproduce

1. Install nuxt-security `npx nuxi@latest module add security`
2. add configuration 
``` 
  security: {
    rateLimiter: {
      driver: {
        name: "redis",
        options: {
          url: "redis://192.168.0.23:6379"
        }
      }
    }
  },
  ```

4. run `npm run dev`
5. see error trying to connect to `127.0.0.1:6379` instead of the url provided in config.

### What is Expected?
Redis attempts to connect to url provided in config

options are not properly passed to unstorage's redis driver. By adding `console.log(opts)
` to `node_modules/unstorage/drivers/redis.mjs` it shows that the redis driver is reciving ```json
{
 driver: 'redis', options: { url: 'redis://192.168.0.23:6379' } }
```

vejja:
Closes #433

## Types of changes

## Description

The Rate Limiter options need to be spread in order to support all the `unstorage` drivers.

Previously, only the `options` property could be provided to the underlying driver, because this driver has only one property with the name `options`.

With this PR, any property name can be provided under the `options` property.

## Checklist:

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -211,12 +211,12 @@ function registerRateLimiterStorage(nuxt: Nuxt, securityOptions: ModuleOptions) 
       securityOptions.rateLimiter ? securityOptions.rateLimiter.driver : undefined,
       { name: 'lruCache' }
     )
-    const { name, options } = driver
+    const { name, options = {} } = driver
     config.storage = defu(
       {
         '#rate-limiter-storage': {
           driver: name,
-          options
+          ...options
         }
       },
       config.storage
--- a/src/types/middlewares.ts
+++ b/src/types/middlewares.ts
@@ -1,16 +1,18 @@
+import type { BuiltinDriverName, BuiltinDriverOptions } from 'unstorage'
 export type RequestSizeLimiter = {
   maxRequestSizeInBytes: number;
   maxUploadFileRequestInBytes: number;
   throwError?: boolean;
 };
 
-export type RateLimiter = {
+export type RateLimiter<T extends BuiltinDriverName = BuiltinDriverName> = {
   tokensPerInterval: number;
   interval: string | number;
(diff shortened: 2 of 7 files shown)


Baroshem (approved):
Thanks for working on it. Unfortunately it will be a breaking change but with a proper release notes, I think we can have it in 2.0.0-rc.1

vejja:
I'm not entirely sure it's a breaking change, it could actually be a bug fix.
It's very difficult to understand how `nitro` sets the `unstorage` options, and then how `unstorage` forwards these options to the underlying drivers. I'll do some local tests and will let you know.

Baroshem:
I think it will be a breaking change purely because the interface for how certain feature will change. Wwith this change, users that uset the previous version of storage options, would have it not working anymore.

But no worries about that. This is not a huge breaking change as long as we provide release notes and migration guide.

vejja:
> I think it will be a breaking change purely because the interface for how certain feature will change. Wwith this change, users that uset the previous version of storage options, would have it not working anymore.
Yes that's right

vejja:
@baroshem : Update after further review
The bug @numselli found out for the `redis` driver was also a problem for all other drivers. The options were actually never passed to the underlying driver.
I'm fixing the TypeScript definitions to clarify how the options must be passed and what values are available for each driver.

Marking as ready now

Baroshem:
Thanks for finding it and fixing it @vejja 💚!

I can't wait for this release. it will be 🔥

Baroshem (approved):

PR #457, opened 2024-05-25
fix: remove navigate-to csp directive

GalacticHypernova:
## Types of changes

## Description

The navigate-to directive was removed from the CSP spec, and is therefore not needed to be configurable. This comments it out so in case a new version of it resurfaces it could easily be added back.

## Checklist:


vejja:
You’re right, thanks @GalacticHypernova for spotting this one

GalacticHypernova:
Glad I could help!

PR #458, opened 2024-05-25
fix(types): allow middleware props to be optional when specified in global config

GalacticHypernova:
## Types of changes

## Description

Allows to specify multiple properties only partially while keeping the rest in default values

## Checklist:


vejja:
I’m not sure they can be optional.
Isn’t it only the case if the default config is used?

GalacticHypernova:
The default config is used regardless in module.ts, so I assume it can be optional?

vejja:
Not if the user deactivates them with ‘false’
Which then might become an issue at route level?

GalacticHypernova:
Well, does the default config per route not work with route level configuration? If a person doesn't specify 'false' at a certain route, won't the default configuration for that route be the one with the default values? If not, then there's the question of whether we want to sacrifice dx for global config to ensure proper values at route level config, vs improving dx for global config at the potential expense of route level config

vejja:
The routes are resolved recursively so if at any level the feature is deactivated with false, all routes nested within that level will not inherit any default values.
Worth double-checking, but I’m pretty sure that this is what defu will do.

Also if the user changes a default value in the global options, he will expect all nested routes to inherit his custom value, not the defaultConfig value. Same for nested paths.

GalacticHypernova:
> Lastly, wouldn't we end up in a situation where we have inconsistent behaviour across features ?

I mean, the throwError field is already optional and I haven't seen any issues with that in any of the middleware?

vejja:
The problem only arises with route-level configurations.
If we want the user to be able to deactivate the feature for a certain route level, then we need to give him the ability to reactivate it for a nested route below. At this point there is no definition of a fallback default anymore.

GalacticHypernova:
Interesting... how about adding specific type overload for per route config that would require all required props?

vejja:
Absolutely, great idea

GalacticHypernova:
I made the route-specific config have all the required props, so now it shouldn't cause any issues.

vejja:
Hey @GalacticHypernova, looks good here
I'm wondering whether I could request 2 additional changes
- merge from 2.0.0-rc.1 instead of main : this is because the `RateLimiter` type has changed
- use only `Required` instead of `RequiredOptional` :  this is because the new `RateLimiter` type in the route rules already eliminates some fields, so for clarity it would be easier to read (I think this request will become obvious when you rebase from 2.0.0-rc.1)

In addition, throwError can remain optional because the implicit default is assumed to be `false`

GalacticHypernova:
> merge from 2.0.0-rc.1 instead of main : this is because the RateLimiter type has changed

I don't think I have access to rebase the PR unfortunately, is there a way you could do it please?

GalacticHypernova:
Updated it to use the RC branch. Anything else you think I should add?

GalacticHypernova:
> `throwError` is now required in route rules, but maybe it could stay optional ?

Done, it should now stay optional.

PR #473, opened 2024-06-04
Update 1.setup.md

hermes85pl:
Provide alternative setting for development in a cleaner way.

## Types of changes

## Checklist:

### DIFF
--- a/docs/content/1.documentation/1.getting-started/1.setup.md
+++ b/docs/content/1.documentation/1.getting-started/1.setup.md
@@ -32,9 +32,11 @@ In order to make this module work with Nuxt DevTools add following configuration
 ```js{}[nuxt.config.ts]
 export default defineNuxtConfig({
   modules: ['nuxt-security', '@nuxt/devtools'],
-  security: {
-    headers: {
-      crossOriginEmbedderPolicy: process.env.NODE_ENV === 'development' ? 'unsafe-none' : 'require-corp',
+  $development: {
+    security: {
+      headers: {
+        crossOriginEmbedderPolicy: 'unsafe-none',
+      },
     },
   },
 });


hermes85pl:
I think that it would be better to find a solution to make DevTools work out of the box, without any additional configuration, as they are becoming omnipresent. In my opinion requiring everyone to do the same change in their configuration should be avoided and the module should resort to providing reasonable defaults instead. Still, until such decision is made, I think that the change that I proposed enables the module's users to make DevTools work in a cleaner way.

Baroshem:
Hey @hermes85pl 

I do agree with your idea. @vejja do you think we should ship such if statement as a part of the default config?

detect what is the environment and if it is dev then set a different value of the header automaically to avoid conflict with the dev tools? We could keep the devtools section with an info that in the default config it is supported but if you change it you can solve the issue like following.

WDYT?

hermes85pl:
So there would be some information that the default config for the development environment is slightly different (with the explanation what exactly and why so), and additionally how to achieve the same (using the above method) to achieve the same in case you provided your own config and the DevTools stopped working for you? Did I get it right?

Baroshem:
Yes, thats the idea so that majority of the users wouldnt have to set the same config in dev environment as majority of the users already use dev tools (we are anning to have some devtools integration as well once I find a time to implement it 😃)

Baroshem:
@vejja what are your thoughts? :)

vejja:
I think it's a good idea !

hermes85pl:
It doesn't fit in my agenda, and this is why I created this PR instead and only suggested a better idea for the final solution.

How do you feel about merging this PR to change the docs like this until there is a better solution?

Baroshem:
I think that we should go for the solution I mentioned so that users would have a better experience out of the box. If you dont have time for developing it, no worries! I will create a PR with these changes soon :)

PR #502, opened 2024-07-31
feat: support server-only (NuxtIsland) components

### LINKED ISSUE #494: nonce is missing when rendering a <NuxtIsland> component during SSR
### Version

nuxt-security: 2.0.0-rc.9
nuxt: 3.12.3

### Reproduction Link

https://stackblitz.com/edit/github-5zfscl?file=pages%2Findex.vue

### Steps to reproduce

Check the network log for the initial document request. You will see that the nonce- is missing from the script-src.

Bildschirmfoto 2024-07-22 um 15 37 11

### What is Expected?

The nonce- should be there. You can remove the `<ServerComponent />` and reload the page. The nonce will be there again.

I added some loggings and it seems like the rendering of the page and the server component share the nonce. The nonce is then removed when rendering the page, because it already exists.

P4sca1:
Closes #494


src/runtime/nitro/plugins/50-updateCsp.ts
```diff
@@ -1,12 +1,19 @@
...
 import { resolveSecurityRules } from '../context'
 import type { ContentSecurityPolicyValue } from '../../../types/headers'
+import { useLogger } from '@nuxt/kit'
 
 /**
  * This plugin updates the CSP directives with the nonce and hashes generated by the server.
  */
 export default defineNitroPlugin((nitroApp) => {
   nitroApp.hooks.hook('render:html', (response, { event }) => {
+    if (response.island) {
```
huang-julien, line 11: Where does `reponse.island` comes from ?

src/runtime/nitro/plugins/40-cspSsrNonce.ts
```diff
@@ -17,18 +18,32 @@ export default defineNitroPlugin((nitroApp) => {
     return
   }
 
+  // Genearate a 16-byte random nonce for each request.
   nitroApp.hooks.hook('request', (event) => {
+    if (isIslandRequest(event)) {
```
huang-julien, line 23: Is it more simple to use routeRules instead of using `.startsWith` ?

src/runtime/nitro/plugins/50-updateCsp.ts
P4sca1: It is provided by Nuxt:

https://github.com/nuxt/nuxt/blob/c0fb30d6d0f06a0c1160e20c9c8f819075d62b14/packages/nuxt/src/core/runtime/nitro/renderer.ts#L503

huang-julien: ah yes 🤦 confused it with the `render:response` hook

src/runtime/nitro/plugins/40-cspSsrNonce.ts
P4sca1: Could you give an example? Comparing the request path is the only way I know to check if a request is for rendering an island component.

huang-julien: We should be able know to augment routeRules

```ts
defineNuxtConfig({
  routeRules: { '/__nuxt_island/**' : {isIsland: true }}
})
```

or we could also mark the event being an island request with an island only plugin (using the env property with the object plugin syntax)

The issue is that this hook gets called at each request so we should make it as performant as possible. Boolean check is always faster than `startsWith()`
image

P4sca1: I don't think that a single `startsWith()` call in the Nuxt request pipeline will have a relevant performance impact.
Using route rules in this case makes the code more complicated in my opinion and requires to augment the TypeScript types correctly.
Route rules need to be matched by `event.path` too, which is likely using a regular expression under the hood, which is slower than a call to `string.startsWith()`.

Baroshem:
Hey there, thanks for this PR! Let's wait for @vejja to take a look at it and afterwards we could merge it :)

Baroshem (approved):

src/utils/island.ts
```diff
@@ -0,0 +1,5 @@
+import type { H3Event } from 'h3'
```
vejja, line 1: My personal preference would be to not use an extra file in  `/utils/`.
If this one-liner is going to be used only in one place, I'd rather see it directly  in the code.

Rationale for this: each time we add a file in `src/utils/`, we need to also augment the `unbuild` entry in package.json, which adds extra maintenance costs. Otherwise the build will fail. My opinion is that `src/utils/` should only contain functions that we need to import in more than one place, of which at least one belongs to the server runtime

src/runtime/nitro/plugins/40-cspSsrNonce.ts
P4sca1: We could instead simply check if the nonce already exist, to not generate it twice. Generating it twice could lead to issues, where the nonce differs at different times of the request pipeline.

src/runtime/nitro/plugins/50-updateCsp.ts
```diff
@@ -1,12 +1,19 @@
 import { defineNitroPlugin } from '#imports'
 import { resolveSecurityRules } from '../context'
 import type { ContentSecurityPolicyValue } from '../../../types/headers'
+import { useLogger } from '@nuxt/kit'
```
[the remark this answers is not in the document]
P4sca1: Will investigate.

src/utils/island.ts
P4sca1: I think the extra entry would not be needed because the isIslandRequest utility is just internal and does not need to be exposed?

However, I agree that a new utility file for a simple function that is only used in one place is not needed. I will update the code and probably remove the check entirely and check for an existing nonce instead.

src/runtime/nitro/plugins/40-cspSsrNonce.ts
P4sca1: I removed the `isIslandRequest` util and replaced it with a simple check if the nonce already exists.

src/utils/island.ts
P4sca1: Done.

src/runtime/nitro/plugins/50-updateCsp.ts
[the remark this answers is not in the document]
P4sca1: It seems like `useLogger` is only available in the module setup function. I changed the code to use `console.warn`. Here is how it looks:

Bildschirmfoto 2024-08-03 um 14 08 10

P4sca1:
@vejja @huang-julien I think I addressed all your concerns. Would be great if you could review again.

vejja:
LGTM
Thank you so much @P4sca1 for this. 
The module now supports islands 🏝️
This is a great addition !

huang-julien (approved):
LGTM :+1:

Baroshem:
Amazing work @P4sca1 

Thank you! I will merge it know to 2.0.0 branch and prepare it for this release :)

P4sca1:
Thank you @Baroshem! Could you draft a new RC release?

Baroshem:
Yes will do that later this week or beginning of the next :)

Baroshem:
@P4sca1 

Yes, sorry but I was on long holidays and couldn't work on it. I expect to release 2.0.0 this week :)

PR #520, opened 2024-09-12
 fix: augment @nuxt/schema rather than nuxt/schema

ThibaultVlacich:
## Types of changes

## Description
Context: nuxt/nuxt#28332

`nuxt/schema` is a re-export of `@nuxt/schema` for users to use. Modules should not augment it, or it may end up overwriting the inferred types from `@nuxt/schema`.

## Checklist:

### DIFF
--- a/src/types/module.ts
+++ b/src/types/module.ts
@@ -38,7 +38,7 @@ export type NuxtSecurityRouteRules = Partial<
   & { requestSizeLimiter: RequestSizeLimiter | false }
 >
 
-declare module 'nuxt/schema' {
+declare module '@nuxt/schema' {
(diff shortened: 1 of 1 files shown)


src/types/module.ts
```diff
@@ -51,12 +51,6 @@ declare module 'nuxt/schema' {
   }
 }
 
-declare module '@nuxt/schema' {
```
Baroshem, line 54: Not sure about this change. zdo you think this should be deleted, not merged with the other change?

ThibaultVlacich: It already was duplicated in the other block :)

Issue #487, opened 2024-07-02
Improve local development with NuxtDevtools


Baroshem:
### Describe the solution you'd like

Modify default config to have a better support for Nuxt Devtools

```ts
crossOriginEmbedderPolicy: process.env.NODE_ENV === 'development' ? 'unsafe-none' : 'credentialless',
```

PR #531, opened 2024-09-30
fix: devtools being blocked in strict mode

dungsil:
## Types of changes

- [X] Bug fix (a non-breaking change which fixes an issue)

## Description

Fix devtools being blocked when strict mode is enabled

88dbb4c4f7d10e05a01336d6bf409e1434aaabd5 the existing documentation has been removed, so it should work the same in strict mode as in none-strict mode 

ref: #487

## Checklist:

### DIFF
--- a/src/defaultConfig.ts
+++ b/src/defaultConfig.ts
@@ -96,7 +96,7 @@ export const defaultSecurityConfig = (serverlUrl: string, strict: boolean) => {
   }
 
   if (strict) {
-    defaultConfig.headers.crossOriginEmbedderPolicy = 'require-corp'
+    defaultConfig.headers.crossOriginEmbedderPolicy = process.env.NODE_ENV === 'development' ? 'unsafe-none' : 'require-corp'
     defaultConfig.headers.contentSecurityPolicy = {
       'base-uri': ["'none'"],
       'default-src' : ["'none'"],


Baroshem:
Hey @dungsil 

Thanks for this PR. @vejja do you think we could change that to allow DevTools to work in Strict? DevTools will work only in development so the code proposed by @dungsil shouldn't affect the production env in any way.

dungsil:
For strict mode, I think it might be a good idea to provide documentation the same as v1.0.  Like adding this to the `Advanced > Strict CSP` document:

vejja:
Hi
Yes this change is ok
By the way we should probably think about refactoring the `Advanced > Strict CSP` section because it is now a source of confusion: the `strict` default config has no relationship with `Strict CSP`.
That section was initially intended as a user guide when we didn't have reasonable defaults that allowed CSP to work in all setups. We have much less issues and questions raised now.

Baroshem (approved):

Baroshem:
Thanks @dungsil @vejja I have merged this PR to 2.1.0 branch that will be used to release a new major version :)

dargmuesli:
I think major would be a 3.x.x, 2.1.x should be a "feature" level release. This being a fix could've translated into a quick 2.0.1 also 😁

Baroshem:
> I think major would be a 3.x.x, 2.1.x should be a "feature" level release. This being a fix could've translated into a quick 2.0.1 also 😁

Correct, I meant minor 2.1.0. Thanks for correcting! :)

PR #539, opened 2024-10-24
fix(core): do not create empty header entries in routeRules

### LINKED ISSUE #538: Vercel deployment bug caused by per-route security config
Hello!

I have a Nuxt project with per-route security config. By deploying it to Vercel it fails with this error:
```
Error: Builder returned invalid routes: should NOT have fewer than 1 properties
```
Local build works without errors, and it seems related to some Vercel-specific things.

I tried to dig in, but it is not yet clear to me why is that happening.
The problem relies on using the `routeRules` security configuration. When I define just some global config of this security module, everything works fine. But when I tried to define it per-route level, Vercel deployment failed with the error noted above. From [this discussion](https://github.com/vercel/vercel/discussions/6623) I understand that it is due to some headers problem, but I'm not sure how the nuxt-security module should be configured properly by that.

### Version

nuxt-security: v2.0.0
nuxt: v3.13.2

### Reproduction Link
[https://github.com/maxdzin/nuxt-security-config-test](https://github.com/maxdzin/nuxt-security-config-test)

### Steps to reproduce
1. Fork specified minimal reproduction repository.
2. Deploy it on Vercel.

### What is Expected?
The Vercel deployment went without any errors and works properly.

By deploying on Vercel the minimal Nuxt + nuxt-security project with per-route security config it fails with this error:
```
Error: Builder returned invalid routes: should NOT have fewer than 1 properties
```

vejja:
## Types of changes

## Description

Fixes #538

The `getHeadersApplicableToAllRessources` was previously returning an empty object `{}` when no eligible entry was found, which resulted in an empty `headers: {}` entry in the routeRules registry.

This could trigger a deployment failure in some Nitro presets (e.g. Vercel).

This PR ensures that no `headers` entry is created in the corresponding routeRule in that case, fixing the bug described in #538.

## Checklist:

### DIFF
--- a/src/utils/headers.ts
+++ b/src/utils/headers.ts
@@ -173,11 +173,16 @@ function appliesToAllResources(optionKey: OptionKey) {
  * Extract the subset of security headers that apply to all resources
  */
 export function getHeadersApplicableToAllResources(headers: SecurityHeaders) {
-  return <Record<HeaderName, string>>Object.fromEntries(
+  const applicableHeaders = <Record<HeaderName, string>>Object.fromEntries(
     Object.entries(headers)
     .filter(([key]) => appliesToAllResources(key as OptionKey))
     .map(([key, value]) => ([getNameFromKey(key as OptionKey), headerStringFromObject(key as OptionKey, value)]))
   )
+  if (Object.keys(applicableHeaders).length === 0) {
+    return undefined
+  } else {
+    return applicableHeaders
+  }
 }
 
 
--- a/src/utils/headers.ts
+++ b/src/utils/headers.ts
@@ -178,11 +178,7 @@ export function getHeadersApplicableToAllResources(headers: SecurityHeaders) {
     .filter(([key]) => appliesToAllResources(key as OptionKey))
     .map(([key, value]) => ([getNameFromKey(key as OptionKey), headerStringFromObject(key as OptionKey, value)]))
   )
-  if (Object.keys(applicableHeaders).length === 0) {
-    return undefined
-  } else {
-    return applicableHeaders
-  }
+  return Object.keys(applicableHeaders).length === 0 ? undefined : applicableHeaders
 }
 
 


maxdzin:
@vejja I tried to make a build of this branch locally (`dev:build`) and it failed with this error:
```
Error: ENOENT: no such file or directory, open './.nuxt/server.headers'
```
Am I missing something?

vejja:
> @vejja I tried to make a build of this branch locally (`dev:build`) and it failed with this error:
> 
> ```
> Error: ENOENT: no such file or directory, open './.nuxt/server.headers'
> ```
> 
> Am I missing something?


I think you are building the playground and hitting the prerenderedHeaders hook. Strange thing that it couldn't write to ./.nuxt/server.headers...
Anyway, just do `prepack` instead to build the module distributable

vejja:
@maxdzin here is the build for the local tests. Let me know !
[dist.zip](https://github.com/user-attachments/files/17511308/dist.zip)

maxdzin:
Nice! Thank you
I tried that and confirm that it working well. The headers passed as expected.

src/utils/headers.ts
```diff
@@ -173,11 +173,16 @@ function appliesToAllResources(optionKey: OptionKey) {
...
+  const applicableHeaders = <Record<HeaderName, string>>Object.fromEntries(
     Object.entries(headers)
     .filter(([key]) => appliesToAllResources(key as OptionKey))
     .map(([key, value]) => ([getNameFromKey(key as OptionKey), headerStringFromObject(key as OptionKey, value)]))
   )
+  if (Object.keys(applicableHeaders).length === 0) {
+    return undefined
+  } else {
+    return applicableHeaders
+  }
```
Baroshem, line 185: suggestion:non-blocking: maybe shorter syntax? :)

```suggestion
  return Object.keys(applicableHeaders).length === 0 ? undefined : applicableHeaders
```

Baroshem (approved):
One small suggestion, apart from that ready to be merged :)

PR #555, opened 2024-11-14
core(fix): Make crypto compatible with Node 18

vejja:
## Types of changes

## Description

This PR re-introduces Node 18 compatiblity.

In v2.1.0, we required a minimum version of Node 20 to be able to run native WebCrypto API functions.

However Nuxt 3 has a minimum requirement of Node 18. As a consequence, Nuxt users who are using Node 18 would not be able to use the Nuxt-Security module.

With this modification we are able to support Node 18, in line with Nuxt 3 minimum requirements.

## Checklist:


Baroshem:
Thanks @vejja I can always count on you! 💚

Issue #564, opened 2024-11-21
SSR nonce injection error: element.replace is not a function with non-string HTML elements


lawren:
### Version

nuxt-security: v2.1.2
nuxt: v3.14.1592

### Steps to reproduce

1. Set up a Nuxt application with Nuxt Security enabled and a Content Security Policy (CSP) that includes a nonce for scripts, styles, and links.
1. Add PrimeVue to the project and use a component like <Button /> or any other PrimeVue component.
1. Run the application in SSR mode (npm run dev or equivalent).

### What is Expected?

The application should render properly in SSR mode, with nonce attributes applied to all script, style, and link tags without errors.

### What is actually happening?

An error occurs during the rendering process:

```
element.replace is not a function
```

This happens because nuxt-security’s `cspSsrNonce.ts` script assumes all elements in `html[section]` are strings, but some elements injected by PrimeVue (or other third-party libraries) may not be strings, causing the replace method to fail.

An argument could be made that this should be fixed on PrimeVue's side, but I figured extra type security doesn't hurt here, and it seems the quickest path to getting everything working again.

[See related issue reported on PrimeVue.](https://github.com/primefaces/primevue/issues/6771)

PR #565, opened 2024-11-21
fix: #564 resolves issue with element.replace on non-string elements

lawren:
## Types of changes

## Description
This PR fixes an issue in the `cspSsrNonce.ts` file where `element.replace` was called on non-string elements in the `html[section]` array during nonce injection. This caused a `TypeError: element.replace is not a function` in SSR mode when third-party libraries, such as PrimeVue, injected non-string elements.

The change introduces a type check to skip non-string elements, ensuring the rendering process doesn’t break and nonce attributes are properly applied to valid elements.

**Why is this change required?**
This resolves a crash in applications using Nuxt Security with SSR and libraries like PrimeVue.

Resolves: #564 

## Checklist:

  - Tests are not applicable in this case since the change addresses runtime behavior with specific third-party library interactions. I did, however, verify existing tests.

### DIFF
--- a/src/runtime/nitro/plugins/40-cspSsrNonce.ts
+++ b/src/runtime/nitro/plugins/40-cspSsrNonce.ts
@@ -52,6 +52,10 @@ export default defineNitroPlugin((nitroApp) => {
     const sections = ['body', 'bodyAppend', 'bodyPrepend', 'head'] as Section[]
     for (const section of sections) {
       html[section] = html[section].map((element) => {
+        // Skip non-string elements
+        if (typeof element !== 'string') {
+          return element;
+        }
         // Add nonce to all link tags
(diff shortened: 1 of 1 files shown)


vejja:
@lawren would you be able to verify if it works now on your side ?
You can use the following stackblitz artefact in your `package.json`: 
```json
{
  "devDependencies": {
    "nuxt-security": "https://pkg.pr.new/Baroshem/nuxt-security@d51c300.tgz"
  }
}
```

lawren:
@vejja verified! Looks good on my end. Really appreciate the quick response on this!

FYI, I just realized the same issue appears to be happening with SRI enabled as well. I am assuming it's the same issue, but just in another nitro plugin. Will try to look into it later today.

PR #567, opened 2024-11-22
fix: #564 resolves issue with element.replace on non-string elements

lawren:
## Types of changes

## Description
This PR fixes an issue in the `subresourceintegrity.ts` file where `element.replace` was called on non-string elements in the `html[section]` array. This caused a `TypeError: element.replace is not a function` in SRI mode when third-party libraries, such as PrimeVue, injected non-string elements.

The change introduces a type check to skip non-string elements, ensuring the rendering process doesn’t break.

**Why is this change required?**
This resolves a crash in applications using Nuxt Security with SRI and libraries like PrimeVue.

Relates to: #564 

## Checklist:

  - Tests are not applicable in this case since the change addresses runtime behavior with specific third-party library interactions. I did, however, verify existing tests.


vejja:
ok

PR #573, opened 2024-11-28
Add basic rate limiter whitelist (specific IPs only)

### LINKED ISSUE #517: Rate limiter whitelist
During load/performances tests, it would be nice to allow defining whitelisted IP (traffic injectors), so there is no need to disable this feature during the tests.

Also this could be interesting to be used when other internal trusted servers are calling the application for batch jobs.

Add a new parameter, a list of IPs that will be whitelisted and not stored in the rate limiter DB, thus allowing no limits for specific hosts.

Disable the rate limiter during the tests

zguig52:
## Types of changes

## Description

Partially resolves #517 (to be discussed further for more advances use cases)

## Checklist:

### DIFF
--- a/docs/content/3.middleware/1.rate-limiter.md
+++ b/docs/content/3.middleware/1.rate-limiter.md
@@ -56,6 +56,7 @@ type RateLimiter = {
   tokensPerInterval: number;
   interval: number;
   headers: boolean;
+  whiteList: string[];
   throwError: boolean;
   driver: {
     name: string;
@@ -82,11 +83,17 @@ The time value in miliseconds after which the rate limiting will be reset. For e
 
 When set to `true` it will set the response headers: `X-Ratelimit-Remaining`, `X-Ratelimit-Reset`, `X-Ratelimit-Limit` with appriopriate values.
 
+### `whiteList`
+
+- Default: `undefined`
+
+When set to `['127.0.0.1', '192.168.0.1']` it will skip rate limiting for these specific IPs.
+
 ### `throwError`
 
 - Default: `true`
 
-Whether to throw Nuxt Error with appriopriate error code and message. If set to false, it will just return the object with the error that you can handle.
+Whether to throw Nuxt Error with appropriate error code and message. If set to false, it will just return the object with the error that you can handle.
 
 ### `driver`
 
--- a/playground/nuxt.config.ts
+++ b/playground/nuxt.config.ts
@@ -42,7 +42,16 @@ export default defineNuxtConfig({
           referrerPolicy: false
         }
       }
-    }
+    },
+    '/rateLimit': {
+      security: {
+        rateLimiter: {
+          tokensPerInterval:1,
+          interval: 1000,
+          whiteList: ['127.0.0.1'],
+        },
+      },
+    },
   },
 
   // Global configuration
--- a/playground/pages/rateLimit.vue
+++ b/playground/pages/rateLimit.vue
@@ -1,0 +1,5 @@
+<template>
+  <div>
+    RateLimit tests Route
+  </div>
+</template>
--- a/src/defaultConfig.ts
+++ b/src/defaultConfig.ts
@@ -55,6 +55,7 @@ export const defaultSecurityConfig = (serverlUrl: string, strict: boolean) => {
       driver: {
         name: 'lruCache'
       },
+      whiteList: undefined,
       ...defaultThrowErrorValue
     },
     xssValidator: {
--- a/src/runtime/server/middleware/rateLimiter.ts
+++ b/src/runtime/server/middleware/rateLimiter.ts
@@ -28,6 +28,9 @@ export default defineEventHandler(async(event) => {
       defaultRateLimiter
     )
     const ip = getIP(event)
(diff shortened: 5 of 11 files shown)


vejja:
Hi @zguig52 
Thanks for this
Do you think it would be possible to add basic tests to `test/rateLimiter.test.ts` to ensure that the module behaves as expected for some basic configuration values? (e.g. undefined, single ip, multiple ips, empty array, etc.)

zguig52:
Hi @vejja ,
I previously added a manual tests page, sorry, I hadn't seen the autotests folder.
I have added 3 pages + associated conf and tests. They all pass when I do yarn test, but I don't know how to get tests results details...
I would have liked to check timing impacts of the whitelist array presence regarding to no whitelist at all.

zguig52:
NB: for the undefined value test, it is already performed with existing tests as it is the default value configured

vejja (approved):

zguig52:
@vejja , this is an IPs CLIENT whitelist that is managed here.

Main usages are for production:
- internal other components calls (where you could whitelist your internal network IPs)
- performance / load tests (where you could whitelist your traffic injectors IPs)

If you look in the function associated to the rateLimiter, it uses client IP to determine and store all calls performed to the endpoint by a specific client and determine if rates are normal:

```
function getIP (event: H3Event) {
  const ip = getRequestIP(event, { xForwardedFor: true }) || ''
  return ip
}
```

This is a h3 function that provide client IP https://github.com/unjs/h3/blob/77510f3c27e6884489189b300c52b5355751aefc/src/utils/request.ts#L297

So I reused this information to compare it to the whitelist and skip rateLimit checks and storage if it matches.

vejja:
> @vejja , this is an IPs CLIENT whitelist that is managed here.
> 
> Main usages are for production:
> 
> * internal other components calls (where you could whitelist your internal network IPs)
> * performance / load tests (where you could whitelist your traffic injectors IPs)
> 
> If you look in the function associated to the rateLimiter, it uses client IP to determine and store all calls performed to the endpoint by a specific client and determine if rates are normal:
> 
> ```
> function getIP (event: H3Event) {
>   const ip = getRequestIP(event, { xForwardedFor: true }) || ''
>   return ip
> }
> ```
> 
> This is a h3 function that provide client IP https://github.com/unjs/h3/blob/77510f3c27e6884489189b300c52b5355751aefc/src/utils/request.ts#L297
> 
> So I reused this information to compare it to the whitelist and skip rateLimit checks and storage if it matches.

Oh oh I think I missed your whole point effectively !

So if I understand correctly, the use cases are significantly larger than what I thought
- For people who deploy with SSR: whenever the SSR server prefetches data from an internal API route, avoid hitting rate limits. Assuming the SSR server is on a known internal IP, then we can whitelist it.
- For larger systems where third-party tools (e.g. testing suites) are used to hit the internal API routes, avoid hitting rate limits. Assuming the 3rd-party suite is a known IP, the we can whitelist.

Am I right here ?

zguig52:
@vejja , that's it! As I started to comment on initial discussion associated, the best would be to be allowed defining networks and not only specific IPs, but it is a bit more complicated to implement and require more thinking!
https://github.com/nuxt-modules/security/issues/517

zguig52:
> npm i https://pkg.pr.new/nuxt-modules/security/nuxt-security@524b003

Just did some tests with this release, it works as expected (at least for a single public IP configured, which is for the moment my tests conditions). I used siege tool to overflow server API routes and it doesn't trigger the rateLimit.

When I am behind the whitelisted public IP, no 429. When I use another public IP, it trigger 429!

vejja:
> @vejja , that's it! As I started to comment on initial discussion associated, the best would be to be allowed defining networks and not only specific IPs, but it is a bit more complicated to implement and require more thinking! #517
> 
> ok clear saw #517 also - would implementation break the type definition of `whitelist` ?


ok clear
saw #517 also - would implementation break the type definition of `whitelist` ?

vejja:
@Baroshem : whenever you think this is ready - I believe this is a feature upgrade, shall we bump to minor v2.2 ?

zguig52:
> saw #517 also - would implementation break the type definition of `whitelist` ?

I will review a bit my comments on this, as I think we might be able to add more capabilities with the same configuration format, without breaking this first implementation.

Baroshem (approved):
Looks great!

Thanks @zguig52 !

@vejja let's plan it for 2.2 release :)

vejja:
Hi @Baroshem 
I think @zguig52 wanted to modify the type definition of `whitelist` to allow more flexible definitions of addresses.
Shall we wait then in order to avoid breaking changes down the line ?

zguig52:
Hi @vejja  and @Baroshem,
From our discussions here https://github.com/nuxt-modules/security/issues/517, I had concluded we are able to keep a simple string array configuration format and then, at startup time, analyse and sort / process internally different kinds of IP/net/ranges definitions!
For me it's all good to go for a first release of this feature, when I will have time, I will work on extending it with more configuration capabilities.

adarshmadrecha:
Any plan to merge soon?

Baroshem:
Hey, yes, thanks for bumping. I will release it in the next days.

PR #593, opened 2024-12-28
fix(nonce): override user-defined nonce values with Nuxt Image

### LINKED ISSUE #586: fix(nonce): duplication
### Environment

```shell
- Operating System: Linux
- Node Version:     v22.12.0
- Nuxt Version:     -
- CLI Version:      3.16.0
- Nitro Version:    -
- Package Manager:  pnpm@9.14.4
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
```

### Nuxt Security Version

v2.1.5

### Default setup used?

### Reproduction

will provide if necessary

### Description

Currently a nonce is added to elements regardless of a nonce being applied already. For example, the `nuxt/image` module allows to set nonces for image preload links. The [regexes](https://github.com/nuxt-modules/security/blob/d46279ae884d5fea4431168a61a112abbd77caea/src/runtime/nitro/plugins/40-cspSsrNonce.ts#L5-L7) could be changed to not match elements which already have a nonce added.

GalacticHypernova:
Resolves #586 
## Types of changes

## Description

For users utilizing Nuxt Image (with its option to provide custom nonces), Nuxt Security will override the user-defined nonce to ensure standard-complying, cryptographically secure nonces.

As a side note, it is worth deprecating the nonce property in Nuxt Image regardless due to the security concerns it brings with no proportionate benefit

## Checklist:

### DIFF
--- a/src/runtime/nitro/plugins/40-cspSsrNonce.ts
+++ b/src/runtime/nitro/plugins/40-cspSsrNonce.ts
@@ -3,6 +3,7 @@ import { resolveSecurityRules } from '../context'
 import { generateRandomNonce } from '../../../utils/crypto'
 
 const LINK_RE = /<link([^>]*?>)/gi
+const NONCE_RE = /nonce="[^"]+"/i
 const SCRIPT_RE = /<script([^>]*?>)/gi
 const STYLE_RE = /<style([^>]*?>)/gi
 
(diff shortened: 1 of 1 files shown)


GalacticHypernova:
I originally thought about making it directly within the link regex, but I couldn't find as of this moment a pattern that is not susceptible to a ReDoS (regular expression denial of service) attack. Therefore, it introduces a lightweight new regex to compare for each found link.

vejja (approved):

vejja:
This one looks ok to me

Baroshem (approved):
Nice work!

PR #615, opened 2025-04-03
fix(sri): resolve hashes using cdnURL at runtime instead of build-time

maxmaxme:
## Types of changes

## Description

This PR fixes a bug in Subresource Integrity (SRI) generation when using dynamic CDN URLs.

Currently, `nuxt-security` uses the `cdnURL` value at **build time** to construct the keys for the `#sri-hashes` map. If the same build is deployed to different environments with different CDN base URLs, the computed `src` / `href` at runtime does not match any entry in the `#sri-hashes`, and `integrity` is not applied.

This change defers the use of `cdnURL` to **runtime**, stripping it from the resource URL before matching it against the `#sri-hashes` map. This makes SRI compatible with multi-environment deployments using different CDN base URLs set via runtime config (`NUXT_APP_CDN_URL`).

### Before:
- `#sri-hashes` keys were generated using `cdnURL` at build time
- runtime matching failed if `cdnURL` changed

### After:
- `#sri-hashes` keys are built without `cdnURL`
- at runtime, actual resource URLs are normalized by removing `cdnURL` before lookup

This allows a single build to work reliably across multiple environments with different CDN domains.

## Checklist:

- [ ] My change requires a change to the documentation. (no)
- [ ] I have updated the documentation accordingly. (no behavior changed)
- [ ] I have added tests to cover my changes (already exists but fails even in the main branch)

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -23,7 +23,7 @@ export default defineNuxtModule<ModuleOptions>({
   },
   async setup (options, nuxt) {
     const resolver = createResolver(import.meta.url)
-    
+
     nuxt.options.build.transpile.push(resolver.resolve('./runtime'))
 
     // First merge module options with default options
@@ -79,7 +79,7 @@ export default defineNuxtModule<ModuleOptions>({
         } else {
           // In case of esbuild, set the drop option
           nuxt.options.vite.esbuild = defu(
-            { 
+            {
               drop: ['console', 'debugger'] as ('console' | 'debugger')[],
             },
             nuxt.options.vite.esbuild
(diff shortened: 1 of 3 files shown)


Baroshem:
@vejja 

Could you take a look at it? I think you have more experience in the SRI part :)

vejja:
Sorry for the delay
This looks alright in principle.
The only things that would make it even more brilliant are:
- there used to be an extra check for whether `cdnURL` was provided with a trailing slash or not. I can remember it arose because of one user reporting the issue. This check has now disappeared, would it still work irrespective of the trailing slash?
- would it be possible to include some basic tests in the test suite to check that it works?
@maxmaxme let me know what you think on these 2

vejja:
Thanks @maxmaxme 
We have another open PR that touches a different aspect of cdnURL : PR #622 - would you mind checking whether it should be backported now into your code?
Also I'm having doubts: when reading the initial code, it looks like when a `cdnURL` option is provided, the `appURL` value is being discarded. In your PR, it looks though that the two values are assumed to be concatenated. Am I misreading something here?

PR #622, opened 2025-04-30
Issue-618: SSG build fails if cdnUrl is relative path

TomHeinemeyer:
If the `cdnUrl` in the Nuxt config is not a valid url, e.g. a relative path like `./` a call to the URL constructor fails because it expects a valid url. In this case just the `cdnUrl` is used.

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -23,7 +23,7 @@ export default defineNuxtModule<ModuleOptions>({
   },
   async setup (options, nuxt) {
     const resolver = createResolver(import.meta.url)
-    
+
     nuxt.options.build.transpile.push(resolver.resolve('./runtime'))
 
     // First merge module options with default options
@@ -79,7 +79,7 @@ export default defineNuxtModule<ModuleOptions>({
         } else {
           // In case of esbuild, set the drop option
           nuxt.options.vite.esbuild = defu(
-            { 
+            {
               drop: ['console', 'debugger'] as ('console' | 'debugger')[],
             },
             nuxt.options.vite.esbuild
@@ -109,7 +109,7 @@ export default defineNuxtModule<ModuleOptions>({
         )
       }
     }
-    
+
     // Register nitro plugin to manage security rules at the level of each route
     addServerPlugin(resolver.resolve('./runtime/nitro/plugins/00-routeRules'))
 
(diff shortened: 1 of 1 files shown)


Baroshem (changes requested):
Hey,

Great work!

Would you be able to fix the linting? The only necessary changes are in lines 332 while others are not needed.

Apart from that, it is ready to be merged :)

vejja:
Are we sure a relative cdnURL is valid?

TomHeinemeyer:
> Are we sure a relative cdnURL is valid?
In our deployment it works and Nuxt also doesn't complain. It is just prepended to the path of the assets, so an actual cdn URL will work as well as local path. In the end the browser will just try to load the assets from whatever we give it and it is our duty to provide a valid path.

TomHeinemeyer:
> Hey,
> 
> Great work!
> 
> Would you be able to fix the linting? The only necessary changes are in lines 332 while others are not needed.
> 
> Apart from that, it is ready to be merged :)
Seems like linting wasn't enabled in my IDE, thanks for spotting this. I will push a fix soon 💪🏼

vejja:
@TomHeinemeyer The official Nuxt docs require an absolute URL https://nuxt.com/docs/api/nuxt-config#cdnurl

I'm not against supporting it if it works for you, however we would need to think carefully about not breaking things further down the line because the SRI hash map is defined relative to `buildAssetsDir` which itself could be based on `cdnURL` (but not always: see https://nuxt.com/docs/api/nuxt-config#buildassetsdir). 
Say you define `cdnURL` as `../foo`, assuming your deployment environment deploys on `https://example.com/bar`, would we still resolve the `sriHashes` correctly ?

Would be really wonderful if you can include some test cases around this and also have a look at #615 to make sure both PRs do not conflict?

TomHeinemeyer:
> @TomHeinemeyer The official Nuxt docs require an absolute URL https://nuxt.com/docs/api/nuxt-config#cdnurl
> 
> I'm not against supporting it if it works for you, however we would need to think carefully about not breaking things further down the line because the SRI hash map is defined relative to `buildAssetsDir` which itself could be based on `cdnURL` (but not always: see https://nuxt.com/docs/api/nuxt-config#buildassetsdir). 
> Say you define `cdnURL` as `'../foo'`, assuming your deployment environment deploys on `https://example.com/bar`, would we still resolve the `sriHashes` correctly ?
> 
> Would be really wonderful if you can include some test cases around this and also have a look at #615 to make sure both PRs do not conflict?


I see, wasn't aware of that. I was just thinking of the deployed application. I will try to make up some test cases for that 👍🏼

PR #658, opened 2025-10-26
fix(cspSsrNonce): more robust tag replacement

### LINKED ISSUE #594: Current cspSsrNonce plugin can result in security problem in some cases
### Environment

```shell
- Operating System: Linux
- Node Version:     v22.11.0
- Nuxt Version:     3.14.1592
- CLI Version:      3.17.2
- Nitro Version:    2.10.4
- Package Manager:  yarn@1.22.22
- Builder:          -
- User Config:      default
- Runtime Modules:  nuxt-security@2.1.5
- Build Modules:    -
```

### Nuxt Security Version

v2.1.5

### Default setup used?

### Security options

```ts

```

### Reproduction

https://github.com/maple3142/nuxt-security-nonce-bug-repro

1. `yarn install && yarn build`
2. `node .output/server/index.mjs`
3. Visit `http://localhost:3000/?statusMessage=1&html=%3Cdiv+id=%22%3Cscript%3E%3Cscript%3Ealert(origin)%3C/script%3E%22%3E%3C/div%3E` from browser

### Description

The browser would pop an alert when visiting that url (so XSS happens here), despite https://github.com/maple3142/nuxt-security-nonce-bug-repro/blob/master/pages/index.vue does sanitize the incoming html using [DOMPurify](https://github.com/cure53/DOMPurify).

The main problem is that nuxt security incorrectly add nonces to script tags using simple string replacement here: https://github.com/nuxt-modules/security/blob/70cd67a4690b53e9167c210e3e020faed901355d/src/runtime/nitro/plugins/40-cspSsrNonce.ts#L53-L73 . And `v-html` can result in unescaped `<script>` to appear in rendered html in SSR mode.

I think nuxt-security should not process html in from v-html, and should not naively use string replacement to do that because it may accidentally replace fake script tags inside element attributes.

### Additional context

_No response_

### Logs

```shell

```

### LINKED ISSUE #610: Script injection vulnerability in combination with useHead
### Environment

```shell
- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.16.0
- CLI Version:      3.23.0
- Nitro Version:    2.11.6
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      compatibilityDate, devtools, modules
- Runtime Modules:  nuxt-security@2.2.0
- Build Modules:    -
```

### Nuxt Security Version

v2.2.0

### Default setup used?

### Security options

```ts

```

### Reproduction

- open https://stackblitz.com/edit/nuxt-starter-94vb5abo?file=app.vue,nuxt.config.ts
- it's a default Nuxt project with default nuxt-security config
- in app.vue, the string `maliciousUserInput` contains two script tags
- this string is used as value for `useHead`/`useHeadSafe` link/meta tags

### Description

- when nuxt-security is turned off, the script is not executed as it's just a string inside of a meta tag attribute value (`<meta property="og:title" content="<script></script><script>alert(1)</script>">`). All good.
- but when enabling nuxt-security, it adds `nonce` attributes which then result in `<meta property="og:title" content="<script nonce="xXNuFdmH0kSn/o6NLkUvBI1S"></script><script nonce="xXNuFdmH0kSn/o6NLkUvBI1S">alert(1)</script>">`; hence it terminates the `content` attribute and the script injection is successful. When opening the page, an alert is shown - which should not happen.
- using `useHeadSafe` only helps partly, it prevents the script injection on the `<link>` tag but the one in `<meta>` tag still persists

On first thought, I see two solutions:

1. make nuxt-security nonce logic smarter to not replace <script> inside of attributes
2. adjust `useHead`/`useHeadSafe` to encode `<>` characters so that nuxt-security's logic won't match anymore

What do you think?

### Additional context

_No response_

### Logs

```shell

```

GalacticHypernova:
## Types of changes

## Description

Closes #594 
Closes #610 
The previous regex implementation, upon further review, had quite a few edge cases. Custom elements, and tags within other elements' props, can cause false modifications, yielding unexpected behavior. This PR fixes the regexes to make them much more robust, fixing said pitfalls. It uses a 2 step approach to mask stringified elements and word-boundaried regex to fix custom elements

## Checklist:

### DIFF
--- a/test/ssrNonce.test.ts
+++ b/test/ssrNonce.test.ts
@@ -112,7 +112,7 @@ describe('[nuxt-security] Nonce', async () => {
     const res = await fetch('/string-script').then(res=>res.text())
 
     const body = res.match(/<div class="(.+)Hello/)
-    const hasNonce = body[1].includes('nonce')
+    const hasNonce = body[1].includes('nonce=')
 
     expect(res).toBeDefined()
     expect(res).toBeTruthy()


GalacticHypernova:
After further testing, it appears as though markup validation regex is not working, and I am not sure why.
Local testing shows that with the markup-validated pattern the resulting chunks are equivalent to the raw ones with the addition of the nonce.

@Baroshem do you perhaps know why that is? Is there some secret configuration in this project that can cause this behavior? Because I statically copied my test page with styles and scripts and compared them in pure JS, and they appear to change correctly.

Besides that, the only thing left for me to figure out is the elements in strings (e.g. other html tags' properties), but we might need a dedicated parser for this... Im trying to look into some kind of html-in-js parser, but it'll take me a while to draft it and experiment with it.

PR #659, opened 2025-11-12
fix(headers): add ssg hashes for `script-src-elem` and `style-src-elem`

dargmuesli:
## Types of changes

## Description
I'm using `script-src-elem` instead of `script-src`, but SSG hashes are only added to the latter and then dropped because `script-src-elem` takes precedence. Thus, my CSP ends up with no script hashes at all.
This PR allows `script-src-elem` to accept hashes aswell and adds a test for that.

## Checklist:


Baroshem (approved):
Amazing work!

PR #672, opened 2026-01-14
feat(csp): add support for Content-Security-Policy-Report-Only header

### LINKED ISSUE #605: Add support for Content-Security-Policy-Report-Only mode
I'm using this library to manage CSP headers, and I need to enable report-only mode. However, the library does not seem to support replacing the Content-Security-Policy header with Content-Security-Policy-Report-Only. This makes it difficult to test policies without enforcing them.

It would be great if the library provided an option to switch to report-only mode, automatically replacing Content-Security-Policy with Content-Security-Policy-Report-Only when enabled.

Manually modifying the response headers after they are set by the library, but this is not ideal.
Using a different middleware for CSP management, but I prefer to keep using this library.

### Additional context
This feature would be useful for safely testing CSP rules before enforcing them in production. Let me know if I can provide more details!

Gonzo17:
## Types of changes

## Description

Support for Content-Security-Policy-Report-Only header, either globally or per route.

Resolves #605 

## Checklist:

### DIFF
--- a/docs/content/1.getting-started/2.configuration.md
+++ b/docs/content/1.getting-started/2.configuration.md
@@ -28,6 +28,7 @@ interface ModuleOptions {
   removeLoggers: boolean | RemoveOptions; // RemoveOptions is being deprecated, please use `true` instead
   ssg: Ssg | false;
   sri: boolean;
+  contentSecurityPolicyReportOnly: boolean;
 }
 ```
 
@@ -119,7 +120,8 @@ security: {
     nitroHeaders: true,
     exportToPresets: true,
   },
-  sri: true
+  sri: true,
+  contentSecurityPolicyReportOnly: false;
 }
 ```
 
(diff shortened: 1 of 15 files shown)


Gonzo17:
> Hey Buddy,
> 
> Thanks for creating this issue. It is clear that a lot of work was done (and a good work)!
> 
> But I am not convinced to this approach with passing report-only in the headers for CSP.
>
> The report only is a completely different header and passing it inside CSPay cause unwanted confusion (as currently everything passed to security.headers.contentSecurityPolicy is converted to native browser headers). This feature will break this stability.

Hey! Thank you for your feedback. :) I'm totally fine with adjusting, I understand your points!

> I think the initial solution with passing it as a separate config option would be better as it does not interfere with the existing native headers :)

I'll update the PR soon. :)

> Also, I highly recommend you to check out this PR -> [nuxt/nuxt#32242](https://github.com/nuxt/nuxt/pull/32242)
> 
> Feel free to contribute that change there as well :)

Oh that's interesting :) Yes, I'd like to do that after finishing this PR!

Gonzo17:
I've pushed the changes, please have a look :)
I also noticed that the FAQ part about CSP Report-Only is redundant now, so we might as well remove it. Just kept it in and updated it accordingly for now.

Baroshem (approved):
Great work! I approve it. Let's wait for few more bugs/features so that I could release a next minor version :)

Issue #699, opened 2026-03-30
Compatibility with nuxt seo 5 ?


GreyXor:
Hello,

I have an client issue with new nuxt-seo version 5.0.

https://github.com/harlan-zw/nuxt-seo/

```
Content-Security-Policy: The page’s settings blocked an inline script (script-src-elem) from being executed because it violates the following directive: “script-src 'self' 'strict-dynamic' 'sha256-ONGmpDWyHYsY3xtj7bHZd+FecK+RbhPN9WDxzTfbRiM=' 'sha256-aO/UyyadceStlstX4piGuLjtRWSu/G7XmmNqJCmbl/8=' 'sha256-45sZv05uwmFdewaAS53saeFyoijT+M2zuwek9asIo/0=' 'sha384-UrS4nC5oMbaktM6vDr57gl5MrqI+1RSPsOLa35iuq12brD/9dRQqdyxcQ9zT7xNj' 'sha384-7lQpU/p0WQUmxZViN4m6v+Bjfrex2NmpULs3qBP8rQN/4cfhNsLjp6TZgQ1YU2h5' 'sha384-nVRKD7S0FoQi/Pm0FQYTpvptOImx/MpAJM3SH4RhjSQclc+Ndq6qp1CpbQwjUTzo' 'sha384-Axg9nYCU5T7QMatdphYsRW1onHfL648Bg/e9kCGJgeg7NZW9WPVSaql+MZFtrsrA' 'sha256-7QIjPOpXT97VD5NmIGqI7WTiAFunWN1i1ifDHVp5i+g='”. Consider using a hash ('sha256-ttsmLGj+N+gpv9yG4IQNFrWKKu5V9rYaWv9tsHam27Y=') or a nonce.
```

```
Content-Security-Policy: Ignoring “'self'” within script-src: ‘strict-dynamic’ specified
  security: {
    headers: {
      contentSecurityPolicy: {
        "connect-src": ["*"],
        "img-src": ["data:", "'self'"],
        "script-src": ["'self'", "'strict-dynamic'", "'nonce-{{nonce}}'"],
        "style-src-attr": ["'unsafe-inline'"],
      },
    },
    strict: true,
  },
```

Was working good with nuxt-seo before 5.0.0

I'm not sure if it's an nuxt-security issue or nuxt-seo issue? 
thanks!

tbshfr:
I have the same problem, here is a simple reproduction: https://github.com/tbshfr/nuxt-csp-repro
[...]

PR #705, opened 2026-04-14
fix: compute CSP hashes from final HTML in render:response

harlan-zw:
## Types of changes

## Description

CSP SSG hashes were computed during `render:html`, so any module that mutated inline `<script>`/`<style>` content in a later `render:html` callback (e.g. a minifier) produced served content that no longer matched the stored hashes, causing CSP violations. `reorderNitroPlugins()` dodged this most of the time but left the ordering fragile.

This moves hash computation, CSP directive update, and CSP meta injection to `render:response`, where `response.body` is the finalized HTML string. Hashes now always match what gets served, regardless of other modules' `render:html` ordering.

Related: [nuxt-seo-utils#103](https://github.com/harlan-zw/nuxt-seo-utils/issues/103).

## Checklist:

- [x] I have added tests to cover my changes

A new fixture plugin (`test/fixtures/ssgHashes/server/plugins/transform-inline.ts`) collapses whitespace inside inline `<script>`/`<style>` tags during `render:html`, modelling a third-party HTML mutator. New parametrized tests in `test/ssgHashes.test.ts` assert every inline script/style in the served body hashes into both the CSP meta tag and the response header, across 8 routes.


GreyXor:
Fix: https://github.com/Baroshem/nuxt-security/issues/699

GreyXor:
@Baroshem Hello, do we need something for a merge ? thanks @harlan-zw !

Baroshem:
Hey folks, sorry for no contact from my side - I was on a longer vacations without access to the computer. I will review it later this week and possibly merge it early next week to unblock you.

Thanks for the patience!

Baroshem (approved):
Great work!

harlan-zw:
Thanks for the review!

PR #674, opened 2026-01-18
fix: use enabled option to not run in nuxt/hints

huang-julien:
## Types of changes

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

https://github.com/nuxt/hints/issues/186#issuecomment-3765429132

following last PR, seems the issue isn't solved yet. I think this option should work no ? 

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes (if not applicable, please state why)

PR #710, opened 2026-05-09
Chore/2.6.0

### LINKED ISSUE #605: Add support for Content-Security-Policy-Report-Only mode

### LINKED ISSUE #684: Default "strict: false" does not work on Nuxt production build
### Environment

```shell
|                      |                                                                                                                                                         |
| -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Operating system** | `macOS 25.3.0`                                                                                                                                          |
| **CPU**              | `Apple M4 Pro (14 cores)`                                                                                                                               |
| **Node.js version**  | `v24.13.1`                                                                                                                                              |
| **nuxt/cli version** | `3.32.0`                                                                                                                                                |
| **Package manager**  | `npm@11.8.0`                                                                                                                                            |
| **Nuxt version**     | `4.3.0`                                                                                                                                                 |
| **Nitro version**    | `2.13.1`                                                                                                                                                |
| **Builder**          | `vite@7.3.1`                                                                                                                                            |
| **Config**           | `alias`, `compatibilityDate`, `devServer`, `devtools`, `eslint`, `ignore`, `imports`, `modules`, `nitro`, `postcss`, `security`, `typescript`, `vite`   |
| **Modules**          | `@nuxt/eslint@1.15.2`, `@nuxt/test-utils/module@4.0.0`, `@nuxt/scripts@0.13.2`, `nuxt-security@2.5.1`                                                   |
```

### Nuxt Security Version

2.1.5

### Default setup used?

### Security options

```ts
defineNuxtConfig({
  security: {
    nonce: true,
    hidePoweredBy: true,
    rateLimiter: false,
    // requestSizeLimiter: false,
    xssValidator: false,
    removeLoggers: false,
  }
})
```

### Reproduction

```
import { setup } from '@nuxt/test-utils/e2e';

await setup({
    rootDir: './packages/frontend',
    dev: false, // <-- this is important to be false
    server: true,
  });
```

### Description

We have Nuxt e2e test which build Nuxt server and our nuxt config does not have "strict" property in security module config meaning that it will be using default that should be "false". But what actually happens is that it will be "true", that revealed through crossOriginEmbedderPolicy header 'unsafe-none' vs. 'credentialess'  https://github.com/Baroshem/nuxt-security/blob/main/src/defaultConfig.ts#L95C27-L95C52

I searched the nuxt-security repository and there are two places where `defaultSecurityConfig()` is called strict true

https://github.com/Baroshem/nuxt-security/blob/094f5644e86d052d3ac7cd94e0ea5a72efcf40c4/src/runtime/server/middleware/rateLimiter.ts#L15

and

https://github.com/Baroshem/nuxt-security/blob/094f5644e86d052d3ac7cd94e0ea5a72efcf40c4/src/runtime/server/middleware/requestSizeLimiter.ts#L8

I expected that when Nuxt builds the app it will build that file only once and the first result will stay. So what I did was setting both `rateLimiter` and `requestSizeLimiter` as `false` and :tada: e2e test was passing and `crossOriginEmbedderPolicy` header value was ´unsafe-none`.

### Additional context

```
import { describe, it, expect, beforeAll, afterAll } from 'vitest';
import { setup, fetch } from '@nuxt/test-utils/e2e';

describe('Server Security Headers for production environment', async () => {
  const fetchHomePage = async (): Promise<Response> => {
    return await fetch('/');
  };

  let originalEnv: string | undefined;
  beforeAll(() => {
    originalEnv = process.env.ENV_NAME;
    process.env.ENV_NAME = 'production';
  });

  afterAll(() => {
    process.env.ENV_NAME = originalEnv;
  });

  await setup({
    rootDir: './packages/frontend',
    dev: false,
    server: true,
  });

  it('has header "Cross-Origin-Embedder-Policy" with value "unsafe-none"', async () => {
    const { status, headers } = await fetchHomePage();

    expect(status).toEqual(200);
    expect(headers.get('Cross-Origin-Embedder-Policy')).toEqual('unsafe-none');
  });
});
```

### LINKED ISSUE #699: Compatibility with nuxt seo 5 ?

### LINKED ISSUE #704: Backtracking regex issue in plugin "cspSsrNonce"
https://github.com/Baroshem/nuxt-security/blob/094f5644e86d052d3ac7cd94e0ea5a72efcf40c4/src/runtime/nitro/plugins/40-cspSsrNonce.ts#L9

This regex can cause a maximum call stack size exceeded issue due to the backtracking pattern on large strings. 

Recommended fix: 

```ts
const QUOTE_MASK_RE = /"([^"\\]*(?:\\.[^"\\]*)*)"/g;
```

Example of error: 

Image

**AI Explanation:** 

The alternation (?:[^"\\]|\\.)* is a catastrophic backtracking pattern. Both branches can match \\., so on a string with no closing quote (or a very long string), the engine tries an exponential number of combinations before giving up. V8 implements regex matching with internal recursion, so on large payloads this exhausts the call stack.

Suggested fix:

Apply the Friedl unrolling technique to eliminate the ambiguous alternation:

This is semantically equivalent but runs in linear time with no backtracking.

Baroshem:
## Description

Resolves https://github.com/Baroshem/nuxt-security/issues/605 Support for Content-Security-Policy-Report-Only header, either globally or per route.
Resolves https://github.com/Baroshem/nuxt-security/issues/704 Backtracking regex issue in plugin cspSsrNonce
Resolves https://github.com/Baroshem/nuxt-security/issues/684 Default strict: false does not work on Nuxt production build
Resolves https://github.com/Baroshem/nuxt-security/issues/699 Compatibility with nuxt seo 5
Update Node to 24.x
https://github.com/Baroshem/nuxt-security/pull/674 https://github.com/nuxt/hints/issues/186#issuecomment-3765429132
Security Advisory GHSA-6cv5-qqmv-j28h
Security Advisory GHSA-7hcp-x3c6-53f3


Baroshem:
The new version will be released in the next few days after I test it properly and make sure that we are not shipping regressions :)

PR #736, opened 2026-07-04
Fix/xss rich text and statusmessage bypass

### LINKED ISSUE #720: XSS validator returns 400 for valid rich-text JSON payload in POST body
## Environment

```shell
- Operating System: Linux
- Node Version:     v24.14.0
- Nuxt Version:     4.2.1
- CLI Version:      3.30.0
- Nitro Version:    2.12.9
- Package Manager:  yarn@1.22.19
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
```

## Nuxt Security Version

v2.6.0

## Default setup used?

## Security options

```ts
defineNuxtConfig({
  security: {
    xssValidator: {
      methods: ['POST'],
    },
  },
})
```

## Reproduction

https://github.com/kouts/nuxt-security/tree/xss_issue_reproduction

## Description

- `xssValidator` rejects a POST JSON payload that contains rich-text-like HTML, even when payload is expected application content.
- In the reproduction branch, test `test/xssValidator.richTextFalsePositive.test.ts` sends:
  - `POST /api/rich-text`
  - body: `{"description":"<p style=\"text-align:center\">hello</p>"}`
- Expected: `200 OK` (payload accepted by endpoint; app-level sanitization can handle content).
- Actual: `400 Bad Request` from middleware.
- Root behavior appears to come from validating `JSON.stringify(body)` as a whole string and failing on any transformed output.

## Additional context

I found this while adding a new tRPC route in my Nuxt app that accepts content from a WYSIWYG editor. The route payload includes rich-text HTML (for example: `{"description":"<p style="text-align:center">hello</p>"}`).

With nuxt-security XSS validator enabled, the request is rejected with `400 Bad Request`, even though this is expected editor content that I sanitize in the app layer before storing/rendering.

As a temporary workaround, I had to manually exclude the affected tRPC endpoints from `xssValidator` in route rules and handle sanitization myself in the route logic. I opened this issue because the middleware behavior is unexpectedly blocking valid rich-text JSON payloads and forced route-level exclusions in a normal editor + API flow.

## Logs

```shell
❯ test/xssValidator.richTextFalsePositive.test.ts (1 test | 1 failed)
× allows a rich-text-like JSON payload sent to POST endpoints

AssertionError: expected 400 to be 200 // Object.is equality
- Expected: 200
+ Received: 400
```

kouts:
## Types of changes

## Description

Resolves: #720

Replaces the XSS validator's stringify-and-compare detection with hook-based malicious payload detection. The old approach compared `xss.process(JSON.stringify(body))` to the input and rejected on *any* difference, which caused false positives for legitimate rich-text HTML like e.g. `<p style="text-align:center">hello</p>`.

The new approach uses the `xss` library's hooks to detect specific malicious patterns:
- Non-whitelisted tags (`<script>`, `<iframe>`, `<svg>`, etc.) via `onIgnoreTag`
- Event handler attributes (`on*`) via `onIgnoreTagAttr`
- Dangerous URL protocols (`javascript:`, `vbscript:`, non-image `data:`) via `safeAttrValue` with an inline regex

Also fixes a security bypass where user-controlled `statusMessage: "Bad Request"` in the request body would skip XSS validation entirely.

Defense-in-depth:
- Recursive payload traversal with depth limit (20) to prevent stack overflow on maliciously nested payloads
- `FilterXSS` instance reused per request instead of per-string
- Regex blocks `data:image/svg+xml` (which can embed scripts) while allowing raster image formats (png, jpeg, gif, webp)

## Checklist:

### DIFF
--- a/src/runtime/server/middleware/xssValidator.ts
+++ b/src/runtime/server/middleware/xssValidator.ts
@@ -1,7 +1,7 @@
 import { defineEventHandler, createError, getQuery, readBody, readMultipartFormData } from 'h3'
-import { FilterXSS, type IFilterXSSOptions } from 'xss'
+import { type IFilterXSSOptions } from 'xss'
 import { resolveSecurityRules } from '../../nitro/context'
-import type { HTTPMethod } from '../../../types/middlewares'
+import { hasMaliciousPayload } from '../utils/xssPayload'
 
 export default defineEventHandler(async(event) => {
   const rules = resolveSecurityRules(event)
@@ -11,38 +11,19 @@ export default defineEventHandler(async(event) => {
       ...rules.xssValidator,
       escapeHtml: undefined
     }
-    if (rules.xssValidator.escapeHtml === false) {
-      // No html escaping (by default "<" is replaced by "&lt;" and ">" by "&gt;")
-      filterOpt.escapeHtml = (value: string) => value
-    }
-    const xssValidator = new FilterXSS(filterOpt)
-
     if (event.node.req.socket.readyState !== 'readOnly') {
-      if (
-        rules.xssValidator.methods &&
-        rules.xssValidator.methods.includes(
-          event.node.req.method! as HTTPMethod
-        )
-      ) {
+      const method = event.node.req.method
+      if (method && (rules.xssValidator.methods as readonly string[]).includes(method)) {
         const valueToFilter =
-          event.node.req.method === 'GET'
+            method === 'GET'
             ? getQuery(event)
             : event.node.req.headers['content-type']?.includes(
                 'multipart/form-data'
               )
             ? await readMultipartFormData(event)
             : await readBody(event)
-        // Fix for problems when one middleware is returning an error and it is catched in the next
         if (valueToFilter && Object.keys(valueToFilter).length) {
-          // Only skip XSS filtering if statusMessage === 'Bad Request' (for error propagation)
-          if (valueToFilter.statusMessage === 'Bad Request') {
-            return
-          }
-          const stringifiedValue = JSON.stringify(valueToFilter)
-          const processedValue = xssValidator.process(
-            JSON.stringify(valueToFilter)
-          )
-          if (processedValue !== stringifiedValue) {
+          if (hasMaliciousPayload(valueToFilter, filterOpt)) {
             const badRequestError = {
               statusCode: 400,
               statusMessage: 'Bad Request'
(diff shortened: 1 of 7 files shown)


src/runtime/server/middleware/xssValidator.ts
```diff
@@ -11,38 +11,19 @@ export default defineEventHandler(async(event) => {
...
                 'multipart/form-data'
               )
             ? await readMultipartFormData(event)
             : await readBody(event)
-        // Fix for problems when one middleware is returning an error and it is catched in the next
         if (valueToFilter && Object.keys(valueToFilter).length) {
-          // Only skip XSS filtering if statusMessage === 'Bad Request' (for error propagation)
-          if (valueToFilter.statusMessage === 'Bad Request') {
-            return
-          }
```
kouts, line 40: Removed this check because `statusMessage` comes from the request body, so it's user controlled. An attacker could send `{ "statusMessage": "Bad Request", "payload": "<script>alert(1)</script>" }` and bypass XSS validation entirely.

If middleware-to-middleware signaling is needed here, `event.context` would be the right mechanism since it can only be set by trusted server code.

src/runtime/server/utils/xssPayload.ts
```diff
@@ -0,0 +1,71 @@
+import { FilterXSS, type IFilterXSSOptions } from 'xss'
+
+const DANGEROUS_URL_ATTRS = new Set(['href', 'src', 'background', 'style'])
+const DANGEROUS_PROTOCOLS = /^\s*(?:javascript|vbscript|data\s*:\s*(?!image\/(?!svg)))/i
+const MAX_DEPTH = 20
```
kouts, line 5: Depth limit here prevents stack overflow on maliciously nested payloads. Returns `true` (conservative) when exceeded, meaning deeply nested payloads are treated as suspicious rather than silently accepted. 20 is generous enough for any real API payload while capping recursion before it can crash...

src/runtime/server/middleware/xssValidator.ts
```diff
@@ -11,38 +11,19 @@ export default defineEventHandler(async(event) => {
       ...rules.xssValidator,
       escapeHtml: undefined
```
kouts, line 12: `escapeHtml` doesn't affect our detection, but the `XssValidator` config type has it as `boolean | undefined` while `IFilterXSSOptions` expects a function type. Explicitly overriding to `undefined` keeps TypeScript happy without an assertion.

PR #738, opened 2026-07-18
refactor(cspSsrNonce): refactor html injection

GalacticHypernova:
## Types of changes

## Description

This PR introduces several changes and fixes issues brought on by #670 and #710 :
1. It unifies the nonce replacement logic from 3 passes (for each elem type) to a single pass, improving performance on large HTML and unifying nonce overwrite/prepending logic (which also means script and style nonces are now deduped)
2. It adjusts the nonce pass to accommodate the lack of outer quotes, so it would capture them correctly
3. **Most importantly, it fixes the XSS openings that the previous PR's re-introduced**

While I couldn't seem to get a minimal test case to demonstrate it, I did encounter it myself in my own app:
image

Where some of the data in NUXT_DATA has gotten malformed:
image

With these changes, the app loads with no issues

## Checklist:

### DIFF
--- a/src/runtime/nitro/plugins/40-cspSsrNonce.ts
+++ b/src/runtime/nitro/plugins/40-cspSsrNonce.ts
@@ -16,9 +16,9 @@ function injectNonceToTags(element: string, nonce: string) {
   const quotes: string[] = [];
 
   // Mask attributes to avoid manipulating stringified elements
-  let maskedElement = element.replace(QUOTE_MASK_RE, (match) => {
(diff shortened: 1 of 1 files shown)


GalacticHypernova:
There may be a way to improve the performance of this fix by converting from a negative look-behind to a regular preceding match (negative look-behind is a slower op), which I tried, but I couldn't get it to work on my test case, so due to the urgency of this PR to patch the existing XSS risks I will keep working on it independently and open a new PR when I find the solution.

